PR #260, opened 2014-10-28
Added ability for callbacks to poll promises for fulfillment value

johnhaley81:
This will allow a callback native to libgit2 to actually have a promise value be returned in the callback eventually. We'll be able to not pause the JS thread while we wait for authentication or any other callback that is triggered from the V8 side.

### DIFF
--- a/generate/descriptor.json
+++ b/generate/descriptor.json
@@ -253,9 +253,6 @@
   "cred": {
     "cType": "git_cred",
     "functions": {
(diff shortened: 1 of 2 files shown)


maxkoryukov:
I still don't understand, why not to use the [Bluebird-lib](http://bluebirdjs.com/)?

johnhaley81:
This PR is superseded by https://github.com/nodegit/nodegit/pull/854

PR #326, opened 2014-12-09
Make remote#download async

tbranyen:


tbranyen:
Okay so the only hold up on this is OS X.  Anyone have a machine they can test with?

johnhaley81:
Confirmed that it is an error on OSX.

```
  101 passing (46s)
  1 pending
  1 failing

  1) Remote can download from a remote:
     Error: No transport bound to this remote
```

maxkorp:
Ruh Roh.
Looking at the diffs in libgit, the error has been updated in 0.22.x to "This remote has never connected". So a transport is not guaranteed to be set by libgit until you've already done a fetch, which sort of defeats the purpose. 

I'm digging a bit deeper now, but we may have to play some games in the C++, but we're gonna have to work in there soon for the iterator stuff, so maybe it's time. Perhaps some sort of generating functions both sync and async with a fallback, or something.

PR #343, opened 2015-01-13
Update dependency fs-extra to ~0.14.0

PeterDaveHello:
Looks like nodegit works fine with fs-extra v0.14, would you like to merge this PR, thanks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
   ],
   "dependencies": {
     "find-parent-dir": "^0.3.0",
-    "fs-extra": "^0.12.0",
+    "fs-extra": "~0.14.0",
     "node-pre-gyp": "~0.5.27",
     "nodegit-promise": "~1.0.0",
     "promisify-node": "~0.1.2"


johnhaley81:
Thanks for bumping fs-extra :)

PeterDaveHello:
Thanks!

PeterDaveHello:
FYI, cdnjs's git auto-updater(https://github.com/cdnjs/autoupdate) not work properly after fs-extra v0.12 (I meant v0.13/0.14), we should still be careful here(I'm not sure if the CI test covered 100% codes).

johnhaley81:
The only thing we actually use fs-extra for is generating the code on a platform that we don't already have a pre-built binary up on S3 and for running tests. So it's not actually used in production at all (hopefully) but just for dev. It should be moved into the dev dependancies but we had some problems in the past with doing that. We'll have to take a closer look at all of our dev dependancies after @maxkorp finishes some of our npm install script work.

PR #350, opened 2015-01-14
WIP: Refactor publishing

maxkorp:
So I've done quite a bit here.
 - First of all, we've minimized the dependency list for normal non dev installs.
 - If the install happens from the NPM registry, we clean up after ourselves (cuts size by ~50%) including some extra dependencies (takes us well over 50% savings).
 - Added the option to do a debug build using BUILD_DEBUG=1, and a shorthand `npm run buildDebug`
 - Moved node-webkit detection to an outside module
 - made it so generate scripts can be called on the command line, but require returns a function, rather than taking immediate effect.
 - moved all scripts out of the root directory and in to lifecyclescripts
 - prepublish step also downloads vendor deps, since those are now published.


maxkorp:
So I'm trying to figure out why this fails on linux 10, something with content type headers not being set...
Definitely pretty confused.

mcollina:
:+1:, but bump the minor version number when you release it. If it introduces the smallest bug, it will trouble people that depends on ~0.2.4. I suggest keeping this new 0.3 release in a new branch, just to simplify life for newcomers to the project, and merge when it is all done.

maxkorp:
I'm fine pushing it to an 0.3.0. Might be time for that bump soon anyways with the libgit changes from upstream.

maxkorp:
This seems to be an issue with openssl. @tbranyen John was saying you solved this before. It seems that on node10 its not finding it.

maxkorp:
During the refactor a version check got missed, which usually caused a different version of http_parser to get pulled in for v10. That wasnt happening, causing a mismatch.

>.<

maxkorp:
Making a new PR for the 0.3 branch.

PR #351, opened 2015-01-15
Updated to nan 1.5.0 and fixed build errors

johnhaley81:
In nan 1.5.0 NanNew does a less-greedy matching for its
overloads so the 2 `New` functions caused a build error so
I renamed the one meant for the js side to `jsNewFunction`.

I'm hoping that's clear enough.


tbranyen:
Woo, is this good to merge ya think?

johnhaley81:
Yeah, I think this is good to merge. I'm curious as to why the appveyor passed this time though. Still trying to figure out why we have it die during clone so often. I feel like it's a timing thing.

PR #362, opened 2015-01-21
update to libgit2 v0.21.5

maxkorp:
new libgit fixes some security issues, so backporting on top of 0.2


maxkorp:
Went ahead and just renamed the json, since the only api change was the addition of one method we don't implement at the moment anyways.

PR #371, opened 2015-01-28
Removes Node 0.11 testing completely

tbranyen:
Doesn't make sense to just remove on Windows.  Lets just drop 0.11 completely for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 env:
   matrix:
     - export NODE_VERSION="0.10"
-    - export NODE_VERSION="0.11"
 before_install:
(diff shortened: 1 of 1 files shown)


johnhaley81:
:+1:

We should look into getting io.js testing into CI.

johnhaley81:
Should we wait on merging this until we have io.js testing in CI?

maxkorp:
Eh, I don't think so. The two are pretty radically different in regards to environment, so until that normalizes i'd just as soon not loop it into CI at all.

tbranyen:
Yeah, we can add in test matrices in the future.  I just don't want to support something that Joyent isn't supporting.

PR #374, opened 2015-01-30
Status and StatusList

orderedlist:
Statusssessssssssss


johnhaley81:
@orderedlist So the latest commit that I pushed is what I think we're going to need to implement to get that callback stuff working. Currently something is breaking the combyne templating when it's loading the partial templates. I saw this in `asyncFunction`, `syncFunction` and `callbackHelper`. @tbranyen might know what's going on but I didn't see anything wrong with the additions.

johnhaley81:
@orderedlist ok this is building now. You should be able to start working with the callbacks now hopefully.

PR #375, opened 2015-01-30
Plaintext credential not async

mbleigh:
I have no idea what I'm doing, but @johnhaley81 told me this would fix a segfault in Gitter :wink:

Issue #323, opened 2014-12-08
2 Step Authentication


johngeorgewright:
A little reminder for people who are using 2 step authentication (I lost a fair while to this :disappointed: ). **You will not be able to use anything other than a SSH connection when using nodegit to operate remotely**. If you're finding that any remote operations are just hanging, this is most likely the cause.

Example:

```js
var nodegit = require('nodegit'),
    clone = nodegit.Clone.clone,
    userpassPlaintextNew = nodegit.Cred.userpassPlaintextNew,
    user = 'my-user-name',
    pass = 'a-very-secret-password', 
    path = '/tmp/somewhere',
    options = {
      remoteCallbacks: {
        credentials: function () {
          return userpassPlaintextNew(user, pass);
        }
      }
    };

function url() {
    var url = 'https://'
    url += encodeURIComponent(user);
    url += ':';
    url += encodeURIComponent(pass);
    url += '@github.com/some-user/some-project.git',
    return url;
}

clone(url(), path, options)
  .then(function (repo) {
    console.log('IF YOU\'RE USING 2 STEP AUTH THIS WILL NEVER HAPPEN!');
  });
```

It's very difficult for automated services to get by 2 step authentication...that's why it's there. I'd recommend using other credentials until there's support for some kind of manual intervention.

maxkorp:
Added an example from this which should be on master soon of how to get this working.

Issue #321, opened 2014-12-05
Needed Examples for 0.3.0


maxkorp:
Here's a list of needed examples. 
 - [x] status (porcelain, z) #170 
 - [x] commit history for a specific file #220 
 - [x] index.addAll #131 
 - [x] 2 Factor Auth
 - [x] Creating a branch

wiky:
is there any examples for 'git status' now?

maxkorp:
Anything that's been implemented will have been checked off, so no examples at the moment. Are you looking for strictly a replica of `git status`?

billt2006:
+1 for `git status`. I will most likely be using something like `git status -s`, but cannot figure out how to get started from the docs. :confused:

dachev:
+1 for git status

maxkorp:
Added 2 factor auth example finally.

johnhaley81:
Status examples are in courtesy of @orderedlist

maxkorp:
https://github.com/nodegit/nodegit/issues/391
Need an example for creating a branch

PR #377, opened 2015-01-30
Make all cred generators sync.

maxkorp:
This was causing weird segfault issues, no idea why any of them were this way (rather, why we didn't override them when we noticed they were async). Also added an example of 2factor auth.

https://github.com/nodegit/nodegit/pull/375
https://github.com/nodegit/nodegit/issues/323
https://github.com/nodegit/nodegit/issues/321


johnhaley81:
Is this good to merge?

maxkorp:
Whoops. Yeah, in it goes.

Issue #380, opened 2015-02-01
Boolean values should evaluate truthy statements


johnhaley81:
Currently if you have a Boolean value in libgit2 you have to send a value of either 1 or 0 which is very awkward in JavaScript. It should determine the truthy-ness of the value passed in and then coerce that into a 1 or 0 which is then sent into libgit2.

maxkorp:
Seems like a natural fit for convert-to-v8-value or whatever it's called.

maxkorp:
Yeah a bit further digging yields 2 components to a solution.
Both require knowing if a function argument or a struct property is a boolean-integer (a 0/1 value).

  1. In `generate/partials/convert-from-v8.cc` if the argument is an int-bool, we can just convert it to a boolean using either `ToBoolean()` (returns a v8 bool type) or `->booleanValue()` (returns an actual C++ bool). 
  1. In `generate/partials/field_accessors.cc`, there are a few options. 
    2. If we want to convert the value permanantly (so if you get it back out of the object, you see the updated value), we can just do that inside the setter in the same way described for function arguments. So the following would hold up.
      
      ```js
      var opts = new nodegit.CloneOptions();
      opts.ignoreCertErrors = {}; // or true or "radishes" or 1
      opts.ignoreCertErrors == 1 // would be true
      ```
    2. If we want to handle maintaining the value the user set (which is how I'm inclined to go), we need to set the actual property (eg git_clone_options' ignore_cert_errors) to the converted value, and maintain a separate client value that can be retrieved.
      ```js
      var opts = new nodegit.CloneOptions();
      opts.ignoreCertErrors = "radishes"; // or true or someObject or 1
      opts.ignoreCertErrors == "radishes"; // (or whatever was passed in) would be true.
      ```
      and in the native side

       ```c++
      myoptionsobject->ignore_cert_errors == 1; // true
      myoptionsobject->ignore_cert_errors_client.getStringValue() == "radishes"// or whatever the syntax is
      ```
    2. We could also just make it show boolean values (so convert it to true/false instead of 0/1), but that feels only a little better than just leaving it, in some ways worse, and definitely no better than the first option.

I'm rooting for the option of maintaining 2 values, even though its a bit more complicated, because it feels more like what a user would expect.

PR #383, opened 2015-02-02
WIP: add test for boolean conversion

maxkorp:
Addresses https://github.com/nodegit/nodegit/issues/380

Started by adding a test. I'll come back to this later in the week.

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -641,9 +641,9 @@ Repository.prototype.fetch = function(
 {
   var repo = this;
 
-  function finallyFn() {
+  function finallyFn(error) {
     if (typeof callback === "function") {
-      callback();
+      callback(error);
     }
   }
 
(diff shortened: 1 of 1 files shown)


tbranyen:
I've been thinking about better monkey patching within our JS-side.  Right now we have a very manual process for normalizing arguments.  I could envision another one called `normalizeBoolean` that simply does a `Number(incoming_value)` transformation to get `true => 1` and `false => 0` before passing onto libgit2.

This feels a bit rough though since it requires manually patching every method we want to make more convenient *by hand*.  I think that's the crap part.  We already scan methods and automatically patch for Promise-support, maybe it'd be useful to try and do this for convenience arguments as well.

maxkorp:
Yeah, I think a lot of this can actually be done in C++ as well, although not strictly necessary.

PR #387, opened 2015-02-03
Fix building when a space is in the path

billt2006:
Building nodegit was breaking if a space was anywhere in the path because `child_process.exec` treats spaces as a delimiter. See http://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback.


johnhaley81:
Thanks for the PR!

PR #390, opened 2015-02-07
Update to libgit2 v0.22.1

johnhaley81:
This is the current status on updating nodegit to use the latest
libgit2. There are some major breaking API changes that need to be
accounted for and will force nodegit to also do a major update as
well.


johnhaley81:
Currently the missing callbacks are breaking the build and that's what I'm working on putting in.

johnhaley81:
Currently playing whack-a-bug with building. Fix a bug/build/fix next bug/build/etc...

johnhaley81:
That test passing on Travis is a false positive.

johnhaley81:
So there's still some things that are funky right now.

With the ["fetch all" test](https://github.com/nodegit/nodegit/blob/3096d0fdd0f37bfffedac9a305fe53ecd3fcae9f/test/tests/remote.js#L137-L151) it's killing nodegit somehow. I think something's going on in C++ that's not getting cleaned up properly. I haven't been able to track it down.

Now with the [http test](https://github.com/nodegit/nodegit/blob/25d7d8887d84d01e6eb1c95e15ec265fa6a202eb/test/tests/clone.js#L35-L43) we have libgit2 anrgy that we're setting a cert check on http since http doesn't use any certification. However, this particular http call redirects to https which then does use a cert which sometimes [dies on osx](http://superuser.com/questions/605900/why-wont-os-x-trust-githubs-ssl-certificate).

I think I'll open up an issue on libgit2 about that since that seems like an oversight.

johnhaley81:
https://github.com/libgit2/libgit2/issues/2904 for HTTP redirect to HTTPS cert check.

johnhaley81:
All of the examples are working. I think after we get Windows green lit we can merge this in!

PR #394, opened 2015-02-10
Updated most dependencies to latest versions

johnhaley81:


johnhaley81:
Updating lodash breaks nodegit. I'll save that one for later.

PR #396, opened 2015-02-10
Update moar dependencies.

maxkorp:
Updates the lodash version. As of 3.0.0, lodash chainable foreach is lazy, and you have to call .value() at the end of the chain for it to evaluate.


johnhaley81:
:+1:

PR #398, opened 2015-02-12
Expose gc so all tests run in CI

tbranyen:
We need to expose garbage collection for all revwalk tests to run, this brings us to 113 passing assertions


maxkorp:
YES!

PR #399, opened 2015-02-12
Made changes to the README

tbranyen:
Added @orderedlist to maintainers, added in badges table, and cleaned up
the "Build from source" section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,29 +3,42 @@ NodeGit
 
 > Node bindings to the [libgit2](http://libgit2.github.com/) project.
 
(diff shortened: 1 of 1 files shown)


tbranyen:
:grey_question:  I'm not completely happy with the table, I don't think it looks as nice.  Anyone else have feedback :question:

johnhaley81:
You don't like the look of the table at all or how it looks inside of the table?

tbranyen:
Both, I know I've seen it done better on other repos, just can't think of which ones.

PR #404, opened 2015-02-14
Tree Entry getBlob() should also support the callback pattern.

jeffwilcox:
Hi,
As this is my first contribution to this project, let me know what I can change, i.e. should I merge to a new branch name or is a direct-to-master preferred. (Contributing.MD didn't have a ton of info on the logistics of moving changes around)

## Change Background

While moving code from the older v0.1.x codebase to the latest, I tried making minimal changes and renames in some functions. A majority of my Node code at this time is not using promises. While calling `getBlob(callback)` on a tree_entry, nothing happened.

I simply mimicked the format used for `getTree` in tree_entry to be consistent (although I'm not setting the `entry` instance on a property like the tree function does).

## Tests

All `mocha` tests passing related to this change. However...

- Clone."can clone with ssh" is a failing test, both before and after the change, on Windows 8.1 x64 with Node v0.12.0.

PR #405, opened 2015-02-14
Tree Entry getBlob() should also support the callback pattern.

jeffwilcox:
While moving code from the older v0.1.x codebase to the latest, I tried making minimal changes and renames in some functions. A majority of my Node code at this time is not using promises. While calling `getBlob(callback)` on a tree_entry, nothing happened.

I simply mimicked the format used for `getTree` in tree_entry to be consistent (although I'm not setting the `entry` instance on a property like the tree function does).

### DIFF
--- a/lib/tree_entry.js
+++ b/lib/tree_entry.js
@@ -63,8 +63,14 @@ TreeEntry.prototype.getTree = function(callback) {
  * Retrieve the tree for this entry. Make sure to call `isTree` first!
  * @return {Blob}
  */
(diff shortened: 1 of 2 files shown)


jeffwilcox:
FYI, this is the rebased version of https://github.com/nodegit/nodegit/pull/404. Not sure why I closed the other PR to do this. I need :coffee: .

tbranyen:
Looks good, thanks!

Issue #406, opened 2015-02-14
segfault when cloning from private BitBucket repo


Lange:
I spoke with @tbranyen about this in Gitter, and he said I should open a formal issue about it.

When attempting to clone from a private BitBucket repo via an HTTPS url, the clone operation will segfault if no `username@` portion of the url is provided AND a `remoteCallbacks.credentials` callback is provided:
```js
var Git = require('nodegit');

// Doesn't log anything, and makes a directory containing nothing but .git in it (no actual content, no errors)
Git.Clone.clone('https://bitbucket.org/endofline/wolv-nowplaying.git', 'tmp', {
    remoteCallbacks: {
        // Never gets called
        credentials: function (url, userName) {
            console.log('I should have gotten called!');
            return Git.Cred.sshKeyFromAgent('langeh');
        }
    }
})
    .catch(function(e) {
        console.error(e)
    });
```

Removing the callback causes the expected `401` error to print.

johnhaley81:
Can you try repo'ing this on #390? That's a pretty big update to the latest version of libgit2 and I'm wondering if still happens after the update.

The code you have above shouldn't need to be modified.

Lange:
I've cloned the current master branch (I noticed that #390 had been merged), and this issue still seems to happen.

This issue pretty low-priority for me, as it's not actually affecting any my production code. I just figured a segfault would be something you'd want to be aware of.

johnhaley81:
Thanks, I do appreciated it. Something like this is definitely going to need to be fixed.

tbranyen:
Yup so the error is that `NanNew` dies on the `NULL` username.   I've added in guards for this. in #419

Lange:
Tested, no segfault! :clap: :beers:

johnhaley81:
#419 is merged. Closing this out.

PR #419, opened 2015-02-24
Fixes segfault issue recorded in #406

tbranyen:
A segfault was occuring when passing null values to `NanNew`, this patch
will the bug for `const char *` and `unsigned int` C types.  We'll
definitely need to add more in the future.

### DIFF
--- a/generate/templates/partials/field_accessors.cc
+++ b/generate/templates/partials/field_accessors.cc
@@ -122,6 +122,20 @@
           return;
         }
 
+        {% each field.args|argsInfo as arg %}
+          {% if arg.name == "payload" %}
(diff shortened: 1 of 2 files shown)


tbranyen:
@Lange if you could give this a branch a shot, I believe it'll address your issue.

Lange:
Tested, no segfault! :clap: :beers:

johnhaley81:
:+1: nice going tbranyen

johnhaley81:
I think the appveyor breaking is unrelated to this PR. I'm going to merge this in and look into fixing the breaking CI in a different PR. Don't want it holding up all of these PR's (#416 and #422).

PR #421, opened 2015-02-24
Allow for saving of props to an object

johnhaley81:
Some properties were being set on the libgit2 C struct but would
be destroyed after the function is completed. The problem was that
C struct would still assume that the memory it had was valid and
would try to use it resulting in random segfaults.

Now you can mark an argument as `saveArg` and then that argument
will be persisted to the parent object and disposed of when the
parent is deleted.

Currently this only works in sync functions since I couldn't find
an async method that would need this. If this changes it shouldn't
be too hard to implement.

Another thing to note is that the if the parent object is marked
as `!selfFreeing` then we're still going to dispose of the
persisted objects because if not then they will memory leak all
over the place. This might need to be looked into more later. I
currently couldn't find a situation where this would happen but
the code is pretty complicated at this point and I certainly could
have missed something

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -1348,6 +1348,13 @@
         "git_remote_rename": {
           "ignore": true
         },
+        "git_remote_set_callbacks": {
+          "args": {
+            "callbacks": {
+              "saveArg": true
+            }
+          }
+        },
         "git_remote_set_fetch_refspecs": {
           "ignore": true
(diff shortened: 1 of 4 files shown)


johnhaley81:
@tbranyen this should hopefully fix #419 and #416

tbranyen:
@johnhaley81 Oh snap, excellent find!

PR #425, opened 2015-02-25
Attempt to fix Windows file locking bug

tbranyen:


tbranyen:
I'm merging this since the tests pass and we fixed the major issues around the file locking.  We are still getting an occasional segfault that we'll need to address in a separate PR.

PR #426, opened 2015-02-25
Fix file lock 2

tbranyen:
testing in CI rando ideas


johnhaley81:
I think that #425 takes care of this. Closing this out.

PR #433, opened 2015-02-27
Skip transfer progress test until it's fixed

johnhaley81:
Currently the transfer progress test is breaking on windows
randomly. @tbranyen and I have been looking into it and it seems
like a libuv issue. I'm going to skip this now so that other
potential PR's aren't held up by it until we can get a good
resolution and fix it in another PR

### DIFF
--- a/test/tests/remote.js
+++ b/test/tests/remote.js
@@ -112,7 +112,7 @@ describe("Remote", function() {
       });
   });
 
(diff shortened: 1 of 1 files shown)


tbranyen:
:+1: No point in getting held up by a random Windows segfault not related to our code.

PR #435, opened 2015-02-27
Fix libuv threading issue

johnhaley81:
`uv_queue_work` is [not thread safe]
(http://docs.libuv.org/en/v1.x/threadpool.html)

This means that all callbacks are in a highly unstable
state and can bring the system down at any time.

Now we're using an [Async handle]
(http://docs.libuv.org/en/latest/async.html#c.uv_async_t)

That should make out callbacks thread safe and consistent
across platforms again

### DIFF
--- a/generate/templates/partials/callback_helpers.cc
+++ b/generate/templates/partials/callback_helpers.cc
@@ -16,7 +16,8 @@
   baton->req.data = baton;
   baton->done = false;
 
-  uv_queue_work(uv_default_loop(), &baton->req, {{ cppFunctionName }}_{{ cbFunction.name }}_asyncWork, {{ cppFunctionName }}_{{ cbFunction.name }}_asyncAfter);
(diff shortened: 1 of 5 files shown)


saghul:
Note that `uv_async_init` is *not* thread-safe. You should initialize it from the loop thread. Only `uv_async_send` is thread-safe.

johnhaley81:
Ahhh interesting. That probably should be a note for that method. I'll fix that in this PR. Thanks for the heads up @saghul!

johnhaley81:
@saghul I put `uv_async_init` on the main loop thread but that seems to kill it in linux/osx.

I opened a new [PR](https://github.com/nodegit/nodegit/pull/438) to show that. Note that Windows succeeds but linux and osx are throwing false positives on the tests.

johnhaley81:
@tjfontaine was saying that we should be doing something like [this](https://github.com/joyent/v8plus/blob/master/v8plus_csup.c#L172-L182).

I think I'm going to merge this for right now and tackle that later.

Issue #401, opened 2015-02-13
Tweak descriptor, run npm install, C++ not updated


tbranyen:
When you tweak the descriptor and run `npm install` the generate scripts do not affect the C++.  If you run `node generate` first the code is updated.  Could you look into this @maxkorp?

maxkorp:
Yeah, I need to find a way aroudn this. The point was to skip the step (and thereby not need to install/load modules) if installing from NPMs registry. I have an idea of how to rework this a bit more now, after talking through an issue on NPMs github a bit.

tbranyen:
I believe this has been resolved as well.

PR #444, opened 2015-03-01
Always regenerate code

tbranyen:
I removed the FS checks that invalidated previously generated source,
since it's so quick to regenerate anyways.

Updated the generate script to exit upon completion.  This resulted in
integrating Promises more heavily.

### DIFF
--- a/generate/index.js
+++ b/generate/index.js
@@ -1,13 +1,27 @@
 var generateJson = require("./scripts/generateJson");
 var generateNativeCode = require("./scripts/generateNativeCode");
 var generateMissingTests = require("./scripts/generateMissingTests");
+var Promise = require("nodegit-promise");
 
 module.exports = function generate() {
-  generateJson();
-  generateNativeCode();
-  generateMissingTests();
+  return Promise.all([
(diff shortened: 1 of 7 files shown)


tbranyen:
Fixes: https://github.com/nodegit/nodegit/issues/401

johnhaley81:
:+1: looks good to me but this might be more up @maxkorp alley.

tbranyen:
Removing this from 0.3.0 milestone.  It's not a blocker at all and can be added in a future 0.3 patch release.

tbranyen:
Luckily that is synchronous atm, so it will always complete first.

maxkorp:
Thinking about this, we actually have an npm script already for building code and rebuilding. I'm gonna do some work in there, what with node-pre-gyp supporting nw-gyp now (and soon atom-shell), but either way I like these changes, other than forcing the generation (that is going to force installing devdependencies even when we download a binary).

johnhaley81:
@maxkorp any progress on this?

maxkorp:
This is included here: https://github.com/nodegit/nodegit/pull/689

PR #445, opened 2015-03-01
Added in an HTTP url for test

tbranyen:
I set up an HTTP backend for my Git repositories.  We can use this for now, but I'd like to have a standalone with GitHub or Bitbucket..  Relying on this URL long-term seems like a bad idea.

### DIFF
--- a/test/tests/clone.js
+++ b/test/tests/clone.js
@@ -25,18 +25,11 @@ describe("Clone", function() {
     });
   });
 
-  it.skip("can clone with http", function() {
+  it("can clone with http", function() {
     var test = this;
-    var url = "http://github.com/nodegit/test.git";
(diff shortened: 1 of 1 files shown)


tbranyen:
This is a valid HTTP repository. Weird that it fails:

```
Error: Invalid Content-Type: text/plain; charset=UTF-8
```

tbranyen:
Turns out I'm not using the smart-http mode, so the headers are wrong.  Once I figure that out, this should pass.

maxkorp:
Github and Bitbucket both force HTTP, unfortunately. We'll HAVE to host some sort of git test server for http :/

tbranyen:
Boo ya!

PR #452, opened 2015-03-02
Add Diff.prototype.findSimilar

orderedlist:
Also DiffFindOptions.

This allows a diff of additions and deletions to be rewritten as renames, or to split a heavily modified file into separate file deletions and additions rather than one modification.


tbranyen:
Nice!

PR #453, opened 2015-03-02
Better cloning with NodeGit

tbranyen:
Right now we have the current call mechanism for clone:

``` javascript
var Git = require('nodegit');

Git.Clone.clone('https://github.com/nodegit/nodegit', 'nodegit').then(function(repo) {
  /* work with the repository */
});
```

This is pretty verbose, compared to:

``` javascript
var Git = require('nodegit');

Git.Clone('https://github.com/nodegit/nodegit', 'nodegit').then(function(repo) {
  /* work with the repository */
});
```

So I've fixed that in the JS-side.  The top-level `NodeGit.Clone` Object is now a Function that behaves like an Object.

This is not a backwards-breaking change, so hopefully we're all in favor?

### DIFF
--- a/test/tests/clone.js
+++ b/test/tests/clone.js
@@ -4,7 +4,7 @@ var promisify = require("promisify-node");
 var fse = promisify(require("fs-extra"));
 var local = path.join.bind(path, __dirname);
 
-describe.only("Clone", function() {
+describe("Clone", function() {
   var Repository = require(local("../../lib/repository"));
   var clone = require(local("../../lib/clone"));
   var NodeGit = require(local("../../"));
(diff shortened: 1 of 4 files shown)


test/tests/clone.js
```diff
@@ -4,9 +4,9 @@ var promisify = require("promisify-node");
 var fse = promisify(require("fs-extra"));
 var local = path.join.bind(path, __dirname);
 
-describe("Clone", function() {
+describe.only("Clone", function() {
```
tbranyen: Damnit!

johnhaley81:
This flows much better than `NodeGit.Clone.clone`

:+1:

tbranyen:
Cool, merging!

PR #454, opened 2015-03-03
StrArray memory fix

mattyclarkson:
We were only allocating enough memory for a `git_strarray` memory pointer rather than the structure.  1a74a2d fixes that.

 7c81612 changes the code to allocate the memory for the structure and the string pointers all in one go to reduce fragmentation. Complete micro-optimisation, happy for it to not get merged but thought I'd do it anyway.


mattyclarkson:
@johnhaley81, could you just put your eyes over this before I merge?

johnhaley81:
This is a much better way to allocate memory for the structure. I added a minor comment but after that :+1:

PR #455, opened 2015-03-03
Include missing lib files in nodegit.js template

orderedlist:
We should probably script these inclusions, so future additions don't get 
missed.


orderedlist:
There were a few classes here (`Reset` for example) that worked in the tests because we were directly requiring the lib file, but not when using `NodeGit.Reset`.

johnhaley81:
Eventually we'll need to do https://github.com/nodegit/nodegit/issues/456

tbranyen:
This now resolves #456 as well.

PR #458, opened 2015-03-03
WIP: Template proto functions

maxkorp:
This makes functions that share typenames available under the type. Eg instead of Clone.clone, you can just call Clone, or Reset instead of Reset.reset, etc.

Also disables the ability to require specific files, requiring you to load nodegit proper.

### DIFF
--- a/generate/templates/templates/nodegit.js
+++ b/generate/templates/templates/nodegit.js
@@ -1,7 +1,5 @@
 var Promise = require("nodegit-promise");
 var promisify = require("promisify-node");
-var path = require("path");
-var local = path.join.bind(path, __dirname);
 var rawApi;
 
 // Attempt to load the production release first, if it fails fall back to the
@@ -54,7 +52,7 @@ exports.__proto__ = rawApi;
 
 var importExtension = function(name) {
   try {
-    require(local(name));
+    require("./" + name);
   }
   catch (unhandledException) {
     if (unhandledException.code != "MODULE_NOT_FOUND") {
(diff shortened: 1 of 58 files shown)


generate/templates/templates/nodegit.js
```diff
@@ -52,19 +54,55 @@ exports.__proto__ = rawApi;
 
 var importExtension = function(name) {
   try {
-    require("./" + name);
-  } catch (unhandledException) {}
+    require(local(name));
```
tbranyen: `local` isn't necessary here, require does this automatically.

generate/templates/templates/nodegit.js
```diff
@@ -52,19 +54,55 @@ exports.__proto__ = rawApi;
...
+  }
+  catch (unhandledException) {
+    if (unhandledException.code != "MODULE_NOT_FOUND") {
```
tbranyen: Good call! This would have trapped errors inside the modules before.

maxkorp: Yeah, You'll never guess how I found that hahaha

generate/templates/templates/nodegit.js
maxkorp: Gotcha. I've just been adding it everywhere in the name of consistency. I was actually having issues with files not loading thinking it wasnt grabbing htem, but they were bing swallowed by the catch on line 59

tbranyen:
I noticed a lot of `require` calls with the `local` normalizing, can you strip those out as well? It'll simplify the code a bit.

maxkorp:
Yeah No problem. A lot of them got added in the tests during the clone fiasco

PR #462, opened 2015-03-04
Fix push example

johnhaley81:


johnhaley81:
Hmm, this shouldn't have died on Appveyor.

johnhaley81:
It's a little worrisome that [1312](https://ci.appveyor.com/project/TimBranyen/nodegit/build/1312) died but [1313](https://ci.appveyor.com/project/TimBranyen/nodegit/build/1313) didn't.

It seems like we might be missing something cleaning up on windows again @tbranyen @mattyclarkson.

mattyclarkson:
Well done for clearing up the example. I was looking through the generated code and it looks like it could do with an audit. I tried putting `valgrind` and `electricfence` around them but it looks like the way node runs stuff they still didn't pick things up. I can look into putting `-fsantize=address` onto the debug builds so that we get a stack trace if anything references out of scope of the allocated memory. The other thing that would be useful is `-fsanitize=thread` as that will catch any threading errors due to the way the generated code is using `libuv`?

Unfortunately the thread and address sanitizers cannot be compiled in together so we'd have to have two build variants. I'd have to look into if that is even supported with `node generate` and `gyp`. It'd be great to have a `Release`, `Debug`, `Address` and `Thread` that instruments the code in certain ways.

johnhaley81:
That's a good idea. I think that would be great for post 0.3.0 release. I'm going to merge this in since it's unrelated.

PR #464, opened 2015-03-05
Don't assign enums to __proto__

orderedlist:
This fixes callstack errors with `apm`. I think the __proto__ assignment is a recent thing, maybe added by @maxkorp? Tests pass without, so not sure the reason for the addition.

What think ye?


tbranyen:
Excellent catch! I was just doing some testing and realized that we're augmenting `Function.prototype` with that `__proto__` assignment :scream: 

``` javascript
> Function.prototype.WALK_MODE
{ WALK_PRE: 0, WALK_POST: 1 }
```

PR #465, opened 2015-03-05
Add automatically generated change log file.

skywinder:
Hi, as I can see, you are carefully fill tags and labels for issues in your repo.
I created a [github_changelog_generator](https://github.com/skywinder/Github-Changelog-Generator), that generate change log file based on tags, issues and merged pull requests (and split them to separate lists according labels) from GitHub issue tracker.
[This Change Log](https://github.com/skywinder/nodegit/blob/add-change-log-file/CHANGELOG.md) was automatically generated by this script.
It helps to easily found what and when was fixed and implemented.
You can easily update this file in future by running script: `github_changelog_generator` in your repo folder.
Hope you find this commit as useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,1049 @@
+# Change Log
+
+## [Unreleased](https://github.com/nodegit/nodegit/tree/HEAD)
+
+[Full Changelog](https://github.com/nodegit/nodegit/compare/v0.2.7...HEAD)
+
+**Closed issues:**
+
+- Suppress astyle errors [\#459](https://github.com/nodegit/nodegit/issues/459)
+
+- io.js support [\#447](https://github.com/nodegit/nodegit/issues/447)
+
+- Meteor: icon fonts not working \(Resource interpreted as Font but transferred with MIME type text/html\) [\#443](https://github.com/nodegit/nodegit/issues/443)
+
+- Don't let users set the payloads through JS [\#436](https://github.com/nodegit/nodegit/issues/436)
+
+- Transfer callback stats are being converted wrong [\#432](https://github.com/nodegit/nodegit/issues/432)
+
+- fetchAll\(\) fails unless a default signature is available [\#431](https://github.com/nodegit/nodegit/issues/431)
+
+- Push example needs to be updated [\#429](https://github.com/nodegit/nodegit/issues/429)
+
+- Question: Is there a certificateCheck option available for pushing to a remote repository? [\#420](https://github.com/nodegit/nodegit/issues/420)
+
+- Repository.open returns empty object [\#412](https://github.com/nodegit/nodegit/issues/412)
+
+- Missing documentation for Tree.walk\(\) [\#411](https://github.com/nodegit/nodegit/issues/411)
+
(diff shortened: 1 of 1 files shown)


johnhaley81:
Any objections to this @nodegit/owners?

maxkorp:
Hrm, I like that a lot for incrementals, not sure on the big update like this one. many of those PRs are fixing things introduced in the release, so it seems odd to include them in the changelog. At minimum, that is an awesome starting point, and potentially an all around solution in many cases (dot releases etc). Awesome awesome tool.

The only other thought is that currently we have history.md, this should be prepended there, either instead of included here, or in addition to (I'm not opposed to having a changelog for the latest version separate from the history).

skywinder:
@maxkorp thanks for mention it. 
It will be a nice feature: to not include PR's, that fix closes another issues. I will research, how to implement it.

About excluding issues: this script can exclude/include issues by specific labels.

maxkorp:
Interesting! That would be great, since we could exclude issues fixed that were introduced in the same release

skywinder:
@maxkorp so, what your opinion: is it better to exclude PR's or issues, that closed by PR from change log?

tbranyen:
@skywinder can you also add a useful hash name to the header anchors?  Something like

``` html
<a name="v0-0-6" href="some_gh_changelog">v0.0.6</a>
```

skywinder:
@tbranyen What do you mean? What should be in place of "some_gh_changelog"?
Now every tag name - is a link to this tag in project: 
For example tag  0.2.4 links to  https://github.com/nodegit/nodegit/tree/v0.2.4

tbranyen:
We've brought this into our site, should we maintain a file in this repo as well?

skywinder:
@tbranyen 
If you want to refer to specific tag in changelog (e.g. Add Anchor Tags To Jump To Specific Location On A Changelog), you can click on the chain image close with tag name:
screen shot 2015-03-17 at 13 17 19

And the browser will redirect you to the link with this tag as anchor:
( https://github.com/skywinder/nodegit/blob/add-change-log-file/CHANGELOG.md#v026-2015-01-20 in that case)
Is it a option, that you are looking for?

maxkorp:
I think it's sort of customary to include the changelog with the source, yeah. We already have HISTORY.md but we can rename it and append it to this.

skywinder:
But now your history has only one tag.
I decide to take this name, because it's kind of standart.  ( keepachangelog.com )

But you can specify your custom name for output file:

    github_changelog_generator -o HISTORY.md

maxkorp:
Yeah, I was saying keep this, and append the contents of history to this ;)

johnhaley81:
:+1:

johnhaley81:
@nodegit/owners how do you guys feel about merging this?

tbranyen:
Yeah might as well.  We can always change it later.  Something is better than nothing.

skywinder:
Hooray, guys! :+1: 
BDW, I will add one feature soon: https://github.com/skywinder/github-changelog-generator/issues/54 
Since you are use quite different labels, rather than default github labels (`Bug` - `bug`, `Modification`- `enhancement `, `Question` - `question`) - this feature will help to make change log even more organised! I'll let you know when I'm done with it!

PR #469, opened 2015-03-05
Change Repository.prototype.setHead to be asynchronous

jrbalsano:
Changes the descriptor file to convert setHead to be asynchronous and
return an error code. Also modifies the createBranch method to
waterfall promises from setHead into Checkout.head.


johnhaley81:
:+1: I'll merge it after tests pass.

PR #472, opened 2015-03-06
Move guides around to remove subindexes

orderedlist:
Makes for nicer browsing.


tbranyen:
Awesome tyvm for doing this!

PR #474, opened 2015-03-07
Add defaults to `Remote.prototype.push`

johnhaley81:
Instead of forcing the user to pass in a signature and message
each time they call push we can derive a good default for them.


tbranyen:
Good call!

PR #475, opened 2015-03-09
Make index.addAll use status to increase performance

maxkorp:
See here: https://github.com/libgit2/libgit2/issues/2687
Doing this speeds things up from around 2500-3500ms for a small-medium repo to around 180-200ms. We should remove this once it's fixed in libgit2, but until then, this seems silly not to add. 

This also adds pathspec.create (since we need that to actually compare pathspecs)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,7 +1,8 @@
 var NodeGit = require("../");
 
 var Index = NodeGit.Index;
-
+var Status = NodeGit.Status;
+var Promise = require("nodegit-promise");
 /**
  * Return an array of the entries in this index.
  * @return {Array<IndexEntry>} an array of IndexEntrys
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -19,7 +21,30 @@ Index.prototype.entries = function() {
...
+    paths.push(path);
+  };
+  var idx = this;
+  return Pathspec.create(pathspec || "*")
+    .then(function(ps) {
+      pathspec = ps;
+      return Status.foreach(repo, statusCB);
+    })
+    .then(function() {
+      return Promise.resolve(paths);
```
tbranyen: Can this just be: `return paths;`?

johnhaley81: Yeah. I don't know the purpose of the `Promise.resolve`

johnhaley81:
Are we ok with putting this in 0.3.0 milestone?

tbranyen:
Works for me.

johnhaley81:
I'm going to take out that `Promise.resolve(paths)` thing first.

PR #476, opened 2015-03-09
Add option to `fetch` to prune the remote afterwards

johnhaley81:
Since pruning a remote is very common after a fetch I figured
having it be an option would be very convenient for the user.


billt2006:
Just noticed there's a missing comma on line 645.
```
repo.fetch(remote, remoteCallbacks, pruneAfter callback)
```

johnhaley81:
What missing comma ;)

billt2006:
I saw nothing...
homer-simpson

PR #479, opened 2015-03-10
Fix atom-shell install on windows

johnhaley81:
The home directory wasn't being set correctly for atom-shell
install in a windows environment. Now things should work again.


tbranyen:
:+1:

PR #481, opened 2015-03-11
Update push example

billt2006:
This should resolve #471.

Issue #471, opened 2015-03-06
Push example is out of date


maxkorp:
helping with #463 I noticed that we have older syntax that isnt supported in 0.3 type code being used in the example for pushing.

PR #483, opened 2015-03-12
Remove unneeded connect call from push example

johnhaley81:


johnhaley81:
Needed to finish closing out #471 and also finishes #481

PR #494, opened 2015-03-18
Guide navigation is currently confusing

tbranyen:
When you click on install basics you get a return to install guides, but
that is actually the root guides.  If you click on install atom shell
and return to install guides it actually brings you back to install
basics.  This is super confusing, I've updated it so that every section
returns to the main list.


tbranyen:
Since you implemented the change originally, do you have any objections with this @orderedlist?

orderedlist:
:shipit:

tbranyen:
:sparkles:

PR #496, opened 2015-03-19
Fix symbolic reference handling in getReferences

billt2006:
I also updated the example in `general.js` so that it would return symbolic references as expected.

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -145,7 +145,12 @@ Repository.prototype.getReference = function(name, callback) {
   }, callback);
 };
 
-Repository.getReferences = function(repo, type, refNamesOnly, callback) {
+Repository.getReferences =
+function(repo, type, refNamesOnly, includeSymbolic, callback) {
+  if (!callback && typeof includeSymbolic === "function") {
+    callback = includeSymbolic;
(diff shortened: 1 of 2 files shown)


lib/repository.js
```diff
@@ -145,7 +145,13 @@ Repository.prototype.getReference = function(name, callback) {
   }, callback);
 };
 
-Repository.getReferences = function(repo, type, refNamesOnly, callback) {
+Repository.getReferences =
+function(repo, type, refNamesOnly, includeSymbolic, callback) {
+  if (typeof callback === "undefined" &&
```
tbranyen: Do we need a typeof check here? That's really only necessary if the variable is possibly undeclared.  Simply do `!callback` or `callback == null`.

billt2006: Alright, I have updated to just use `!callback`.

tbranyen: Awesome, I'm being picky here as someone could pass a falsy value that isn't undefined and that would break your check.

billt2006: No problem! You have every right to be picky. :wink:

johnhaley81:
:+1: looks good. I'm going to merge this.

PR #502, opened 2015-03-22
Improve coverage

tbranyen:
This removes files that we didn't author from our coverage report (nan, gyp, etc).  I removed some dead code from `src/functions/copy` and then added some Note tests to boost up some 0 coverage files.

### DIFF
--- a/generate/templates/manual/src/functions/copy.cc
+++ b/generate/templates/manual/src/functions/copy.cc
@@ -9,39 +9,3 @@ const git_error *git_error_dup(const git_error *arg) {
   result->message = strdup(arg->message);
   return result;
 }
-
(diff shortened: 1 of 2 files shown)


test/tests/note.js
```diff
@@ -0,0 +1,52 @@
...
+      assert.equal(note.message(), "Testing!");
+    });
+  });
+
+  it("can be removed", function(done) {
+    var sha = this.commit.id();
+    var noteRef = "refs/notes/commits";
+    var sig = Signature.create("John", "john@doe.com", Date.now(), 0);
+
+    Note.remove(this.repository, noteRef, sig, sig, sha);
```
tbranyen: This should definitely be async.

PR #504, opened 2015-03-22
Add note tests

tbranyen:


tbranyen:
FWIW these work locally on my machine, not sure what's up...

PR #524, opened 2015-04-02
Add convenience methods to status file

maxkorp:
Adds methods `inWorkingTree` and `inIndex` to StatusFiles. These just check it the status contains any WT_xxx status codes (or INDEX_xxx). It's very possible that a status could have both (eg if you've done a patch stage).

### DIFF
--- a/lib/status_file.js
+++ b/lib/status_file.js
@@ -82,6 +82,20 @@ var StatusFile = function(args) {
     },
     isIgnored: function() {
       return data.statusBit & codes.IGNORED;
+    },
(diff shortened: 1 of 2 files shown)


maxkorp:
My only thought is that I'm not super hot on the method names. Anybody have any better ideas?

johnhaley81:
I think the names are ok. Maybe `isInWorkingTree`? But that's sounding to verbose to me.

tbranyen:
The names make sense to me.  Merging, thanks!

Issue #501, opened 2015-03-22
Bug with oid implicit cast inside C++


tbranyen:
Found a segfault when using `this.commit.sha()` inside a `Note.remove` test.  I switched over to `this.commit.id()` instead and the segfault disappeared.  The error I received was:

```
*** Error in `/usr/bin/node': free(): invalid size: 0x00007fffdbe986c0 ***
======= Backtrace: =========
/usr/lib/libc.so.6(+0x7198e)[0x7f802fc2198e]
/usr/lib/libc.so.6(+0x76dee)[0x7f802fc26dee]
/usr/lib/libc.so.6(+0x775cb)[0x7f802fc275cb]
/home/tim/git/nodegit/nodegit/build/Release/nodegit.node(_ZN7GitNote6RemoveERKN2v820FunctionCallbackInfoINS0_5ValueEEE+0xb42)[0x7f802ccde772]
/usr/bin/node(_ZN2v88internal25FunctionCallbackArguments4CallEPFvRKNS_20FunctionCallbackInfoINS_5ValueEEEE+0xa7)[0x5f8ab7]
/usr/bin/node[0x61590c]
[0x22e51d20740e]
```

johnhaley81:
Fixed in #531

PR #531, opened 2015-04-07
Fix freeing a `GitOid` that was passed as a string

johnhaley81:
We will allow a user to pass a string as a valid parameter for any parameter that is a `GitOid`. In the function will will then convert that parameter to a `GitOid` and then free it after it's no longer in use. In a synchronous function we just did `free(&arg)` which was causing an issue in Windows. Now we'll appropriately cast that to a `void *` that will allow MSBUILD to free it without crashing.


tbranyen:
Pretty sure this fixes: https://github.com/nodegit/nodegit/issues/501 right?

johnhaley81:
That should fix it.

johnhaley81:
Ugh, timeouts are killing the tests...

PR #532, opened 2015-04-07
Bump timeout on `remote.download` test

johnhaley81:


tbranyen:
I think sometimes the tests will just fail.  It sucks and indicates a bug somewhere, but I'm not sure if bumping timeouts is the fix =/

Oh snap the timeout it died on is initializing a new repo.  That could be FS timeout.  It sucks, but these have always really been integration tests.  It may make sense to bump the timeout globally.  You'll notice it died on 2sec, we can override the global default and make it 18sec or so.

maxkorp:
This is not really worth keeping around at this point, and can be revisited if needed.

Issue #533, opened 2015-04-09
`options` not reusable, nodegit destroys it


callumacrae:
https://github.com/nodegit/nodegit/blob/master/lib/clone.js#L21

Why are you doing this? This prevents reuse, e.g. I want to clone, and then fetch.

```js
let remoteOpts = {
	remoteCallbacks: {
		certificateCheck: function () {
			return 1;
		},
		credentials: function (url, username) {
			return git.Cred.sshKeyFromAgent(username);
		}
	}
};

function getFileFromBranch(branch, filePath) {
	let repository;

	return git.Repository.open(config.git.cloneTo)
		.catch(function (err) {
			if (err.message.startsWith('Failed to resolve')) {
				winston.info('Repo not found, cloning');

				// USED HERE
				return git.Clone(config.git.repo, config.git.cloneTo, remoteOpts);
			} else {
				winston.error(err);
				throw err;
			}
		})
		.then(function (repo) {
			repository = repo;

			winston.log('verbose', 'Changing branch');
			return repo.checkoutBranch(branch);
		})
		.catch(function (err) {
			if (/Reference '([^']+)' not found/.test(err.message)) {
				winston.info('Branch not found, fetching');

				// USED HERE
				let remoteCallbacks = remoteOpts.remoteCallbacks;
				return repository.fetch('origin', remoteCallbacks, true)
					.then(function () {
						return repository.checkoutBranch(branch);
					});
			} else {
				winston.error(err);
				throw err;
			}
		})
		.then(function () {
			return fs.readFile(path.join(config.git.cloneTo, filePath));
		});
}
```

tbranyen:
As you can see only a few lines further down, we reattach it with the options normalized:

https://github.com/nodegit/nodegit/blob/master/lib/clone.js#L27-L28

Your code should work just fine.  If not please let us know what's broken.

callumacrae:
My code was throwing an error because `remoteOpts.remoteCallbacks` was undefined. Not sure why, but it isn't my code!

tbranyen:
@callumacrae ah I see what the issue is.  We normalize the entire options object thus overriding the one you passed.  I'm not entirely sure why we have the `delete` there anymore... I'll try removing it and see if the tests still pass.

tbranyen:
Created a PR: https://github.com/nodegit/nodegit/pull/534

tbranyen:
Yup, here is the bug that is introduced by removing the `delete` keyword:

[...]

PR #534, opened 2015-04-09
Attempt remove the delete keyword

tbranyen:
Resolves https://github.com/nodegit/nodegit/issues/533


johnhaley81:
Do we know why this is failing?

tbranyen:
It was failing because `normalizeOptions` doesn't know to exclude `remote_callbacks`, I could have hardcoded it there, but I preferred to just take a shallow clone of the passed in options object as to not mutate the one passed.

PR #545, opened 2015-04-20
Add get/set config string methods and tests

johnhaley81:


s-xiang:
Got a question about the Config, is there any way to read all the variables under the user? 
For example, getString("user.email") will return the user email, can I get a list of variables under [user] instead of just email or name? Thanks.

Issue #561, opened 2015-05-11
Globally installed NodeGit wipes out NPM


tbranyen:
http://stackoverflow.com/questions/30128359/running-npm-install-unbuilds-global-npm

@maxkorp I know you wrote the cleanup code, any idea if we uninstall NPM? If so, we probably should stop doing that :-p

andrepadez:
oh, thanks @tbranyen . Just came here to open this issue...

maxkorp:
FYI if this _is_ an issue, the no more cleaning thing that landed yesterday should fix it.

andrepadez:
I started using 0.4.0 today, and i still get npm unbuilt...

maxkorp:
Yeah, that's landed in master, but there's on more critical bug waiting to clear (fix is found) before we publish the update.

andrepadez:
cool thanks. let us know when, please :+1:

tbranyen:
@andrepadez no specific time frame, you can install from master with `nodegit/nodegit#master` in your package.json.

andrepadez:
thank you @tbranyen . will do

andrepadez:
actually not. 
I run npm install on my module, it installs nodegit in my node_modules, but when i run my project node says it can't find module 'nodegit'.

I checked the package.json where the "main" is "lib/nodegit.js" and that file is not there.
I tried changing to "lib/index.js" but it throws some error so probably that shouldn't be the entry file also.

Help please?

tbranyen:
@andrepadez sounds like a compile failure.  What OS are you running?

andrepadez:
OSX Yosemite 10.10.2 (14C2055)

tbranyen:
Hrm I know @johnhaley81 and @maxkorp are running OS X, maybe they can chime in if there's anything special you need to do before the compile will work.  I'd imagine you have XCode installed right?

andrepadez:
@tbranyen yes i do

johnhaley81:
I've been building 0.4.0 on OSX 10.10.3 in node 0.10.38 and 0.12.2. Are you trying to install in nw.js or electron by chance?

andrepadez:
here's my usecase:
```
$ git clone git@github.com:vigour-io/gaston.git
$ cd gaston
$ git checkout new-gaston
$ npm install
$ ./bin/gaston.js
```
with this, i get module nodegit not found.

@johnhaley81  Can you try it? thanks

johnhaley81:
[...]

PR #562, opened 2015-05-11
stop cleaning on post-install

maxkorp:
Lots of issues stemming from cleanup. Useful, so I've left the script, but it no longer runs automatically now.


maxkorp:
_should_ fix https://github.com/nodegit/nodegit/issues/561

tbranyen:
Word, thanks for addressing this quickly @maxkorp!

PR #566, opened 2015-05-12
Fix build issues with 0.4.0

### LINKED ISSUE #565: Build is broken on windows
Getting lots of libssh issues during compile, seems tied to the commit where libssh was updated from 1.4.3 to 1.5.0. Trying a rollback now.

maxkorp:
Fixes #565 
Waiting to merge until we can get the false positives resolved

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "version": "0.22.2"
   },
   "libssh2": {
-    "url": "http://www.libssh2.org/download/libssh2-1.5.0.tar.gz",
-    "version": "1.5.0"
+    "url": "http://www.libssh2.org/download/libssh2-1.4.3.tar.gz",
+    "version": "1.4.3"
   },
   "http_parser": {
     "url": "https://github.com/joyent/http-parser/archive/v2.4.2.tar.gz",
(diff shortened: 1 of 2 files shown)


johnhaley81:
@maxkorp I think the false positives in CI should be a different PR. We should get this merged in ASAP IMO.

package.json
```diff
@@ -70,7 +70,7 @@
     "lcov-result-merger": "^1.0.2",
     "lodash": "^3.8.0",
     "mocha": "^2.2.4",
-    "nan": "~1.7.0",
+    "nan": "~1.8.0",
```
tbranyen: Isn't latest NaN, 1.8.4? If that works, can we change this to `^1.8.4`?

johnhaley81: Yeah, np. It was downloading 1.8.4 but we should up that regardless.

johnhaley81:
Seg faults in iojs on linux :(

maxkorp:
I'd like to get it working on iojs in linux before merging, but I'm cool waiting on the false positives thing. I can take a crack at it in a few.

johnhaley81:
I'm unable to repo the seg fault on linux. @tbranyen could you give this a try?

maxkorp:
So apparently its an issue with iojs 2+. Any ideas/thoughts?

johnhaley81:
Only in linux, OSX looks to be good. Have you tried windows?

johnhaley81:
Removed the bumping nan and supporting io.js from this PR.

Issue #570, opened 2015-05-15
tree_entry path function not working when calling getEntry with a path including subdir


jfremy:
When building the path of a tree entry, nodegit recursively take the path of each entry's tree parent.
However this does not work if the entry was in a subdir of the parent tree because then we'll be missing the subdir.

As an example:
getEntry("test/rawcommit.js") => entry path will be returned as "rawcommit.js"

 I'm submitting a push request + test with a potential fix for discussion

PR #571, opened 2015-05-15
Fix path function in tree_entry (fix for issue #570)

jfremy:
Fix for issue #570
If the entry path goes beyond the immediate children (a file in a subdir for instance) then path would return a path missing the subdir name

As an example:
getEntry("test/rawcommit.js") => entry path will be returned as "rawcommit.js"

### DIFF
--- a/lib/tree.js
+++ b/lib/tree.js
@@ -1,3 +1,4 @@
+var pathmod = require("path");
 var events = require("events");
 var NodeGit = require("../");
 var Diff = NodeGit.Diff;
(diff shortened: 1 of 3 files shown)


lib/tree.js
```diff
@@ -1,3 +1,4 @@
+var pathmod = require("path");
```
tbranyen: Can we rename this `path` and change the variable later on to `filePath`?

tbranyen:
Thanks for this, one minor request!

tbranyen:
Looks good, tyvm!

PR #577, opened 2015-05-18
Check for existence to avoid throwing an error when there is no default signature

tomruggs:


tbranyen:
Looks good, think you could write us a test as well? If you need help hop in the Gitter channel and we'll help ya with it :D

tomruggs:
The existing test has been fixed to fail on error.  I verified that the test failed without my fix and passes with it.

johnhaley81:
Thanks for the PR!

tomruggs:
Np.  Thanks for the merge.

PR #579, opened 2015-05-19
Added NodeGit.Checkout.index

jdgarcia:
+ updated `descriptor.json` for `git_checkout_index` (was previously ignored)
+ added normalize option convenience method
+ added a test to make sure an index with conflicts will write to disk when checked out

This function is useful for easily writing merge conflicts to disk since both merge and rebase will throw the index when there are merge conflicts


johnhaley81:
Thanks man. This looks great!

PR #580, opened 2015-05-19
Fix encrypted SSH keys

### LINKED ISSUE #553: NodeGit.Cred.sshKeyNew: Waiting for USERAUTH response
Hey guys I am trying to fetch/clone using *Cred.sshKeyNew*, but I keep getting: 

``Failed to authenticate SSH session: Waiting for USERAUTH response``

Source? https://github.com/karelia/libssh2/blob/master/src/userauth.c

Does anyone know whats going on?

I am using just like https://github.com/nodegit/nodegit/blob/master/test/tests/clone.js#L92-L114

I can get everything to work by using *NodeGit.Cred.sshKeyFromAgent(userName)*. But in my case, I will have to make it work using a private key file.

johnhaley81:
This should fix #553


johnhaley81:
Currently a WIP.

lucastschmidt:
Hey @johnhaley81 , I tried to install your fix branch to test it here, but I wasnt successful. 

Basically, I noticed that the latest version *0.4.0* cant be installed here on Mac. I dont know whats up, but after the install is successful, the require('nodegit') fails, since there is no *./lib/nodegit.js* file here. I dont know whats going on. It seems that the file is not created here. The only *nodegit.js* file found in my project is the one from the template folder.

Do you have any idea what I may be doing wrong?

johnhaley81:
So first things first, this is "passing" tests even though [it's failing on CI](https://travis-ci.org/nodegit/nodegit/jobs/63105686#L923). Don't know why this is happening but it's very wrong.

@lucastschmidt So you need a proper build setup to compile nodegit from source. We have [this guide](http://www.nodegit.org/guides/install/from-source/) which you can use to figure out what you're missing.

Also, this just added a failing test so I could repo your issue with #553. Now it's not an issue with the code but rather an issue that I added a key to use in the tests but never added it to the server. As soon as I add it to GitHub the test will work fine.

johnhaley81:
So the main issue with this is that we don't have an account registered with the provided SSH key. As soon as we can do that the test should pass and we can merge this in.

maxkorp:
Rebased, fixed it to match the current api, and added the key to have read-only access to the test repo :+1:

PR #582, opened 2015-05-19
Fix CI false positives

johnhaley81:


johnhaley81:
Windows is still broken, that will be fixed in #566

PR #589, opened 2015-05-22
Standard merge

jdgarcia:
Enabled `git_merge` (NodeGit.Merge() and NodeGit.Merge.merge) and added a convenience method.

`git_merge` is the standard merge that merges into HEAD and writes the MERGE_* to the git repo which is the indicator that the repo is currently in a merge state. (`git_repository_state` looks for those files to determine if it's in a merge state)

Other merge methods, such as `Merge.commits` don't actually leave the repo in a merge state

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -972,7 +972,14 @@
     "merge": {
       "functions": {
         "git_merge": {
-          "ignore": true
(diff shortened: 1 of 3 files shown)


jdgarcia:
"hey, tests passed"

PR #590, opened 2015-05-28
Rewrite test for statusFile

vanadium23:
I'm adding test to check all Statuses, that have NodeGit.Status object. Currently, I skip statuses 'CURRENT' and 'WP_UNREADABLE', because it fails tests. Are they needed to implement different logic, then in checkStatusFile function?

### DIFF
--- a/test/tests/status_file.js
+++ b/test/tests/status_file.js
@@ -7,37 +7,79 @@ describe("StatusFile", function() {
 
   var pathName = "README.md";
 
-  var checkStatusFile = function(status) {
-    it("identifies the proper statuses for " + status, function() {
-      var statusFile = new StatusFile({
-        path: pathName,
-        status: Status.STATUS[status]
-      });
-      var specialFunction = status.replace(/^(WT|INDEX)_/, "");
-      specialFunction = "is" +
-        specialFunction[0] +
-        specialFunction.substring(1).toLowerCase();
-      if (/^WT_/.test(status)) {
-        assert.ok(statusFile.inWorkingTree());
-        assert.ok(!statusFile.inIndex());
-      }
-      if (/^INDEX_/.test(status)) {
-        assert.ok(!statusFile.inWorkingTree());
-        assert.ok(statusFile.inIndex());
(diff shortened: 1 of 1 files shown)


tbranyen:
@vanadium23 I'm a bit hesitant with this change.  I'd favor explicitness in the test code, as it introduces less chances for errors and makes the tests generally easier to read and tweak.  For instance the statuses you're skipping: https://github.com/nodegit/nodegit/pull/590/files#diff-2d1e0757a1b22fb3b45bd37e8fd61ca4R36 could be expressed with `it.skip`.

Just my thoughts on this.

vanadium23:
On one hand, I'm agree, that tests needs to be as straight as possible, but in this case with very small piece of logic we can check all statuses and view error not only for files in WT_NEW. All in all, to merge or not to merge is on your own. Thanks for response.

tbranyen:
@vanadium23 nah its not just my decision as there are quite a few maintainers.  I was just voicing my opinion.

johnhaley81:
TBH I'm torn. It's a pretty nice function that just goes through and confirms that we can grab statuses out of the repo. However I really like to adhere to being as explicit as possible with tests. Helper functions to setup and teardown scenarios and whatnot but not to construct the tests themselves.

maxkorp:
Agreed. Definitely would prefer individual tests, but definitely a +1 for testing more statuses (statii?). Plus even skipped, the status tests for 'CURRENT' and 'WP_UNREADABLE', would be super awesome.

vanadium23:
@maxkorp Add individual test with helper function. Also skips for statuses 'CURRENT' and 'WP_UNREADABLE'.

tbranyen:
:+1:

maxkorp:
Awesome! :smile:

vanadium23:
Is there anything else you need in this pr?

tbranyen:
Nope seems good to me.  I'll merge in.

maxkorp:
Scooore. Thanks @vanadium23!

Issue #591, opened 2015-05-29
TreeEntry has no dirtoparent when it is obtained via Tree.entries()


tomruggs:
I'm iterating over Tree.entries() and working with each individual TreeEntry.  The problem is that if I attempt to call toString() or path() on the TreeEntries I get:

```
[90m    [36mTypeError: Arguments to path.join must be strings
        at path.js:360:15
        at Array.filter (native)
        at Object.exports.join (path.js:358:36)
        at TreeEntry.path (/opt/packages/teamcity/work/739e0919cda857a9/node_modules/nodegit/lib/tree_entry.js:82:15)
```

This is because the TreeEntries that come from the entries() list do not go through getEntry() and therefore do not ever have a dirtoparent property assigned to them.  

This was introduced a few weeks ago here: 

https://github.com/nodegit/nodegit/commit/c7f86de01542e75825f7e537c7725ab374ab6ffd#diff-a170a8f0bb4e54cd4900b5fc814eeccdR82

johnhaley81:
This is fixed in #592

PR #592, opened 2015-05-29
Fix for issue #591.  TreeEntry.path() throws when TreeEntry came from Tree.entries()

tomruggs:


package.json
```diff
@@ -74,7 +74,8 @@
     "nw-gyp": "^0.12.4",
     "pangyp": "^2.1.0",
     "request": "^2.55.0",
-    "tar": "^2.1.0"
+    "tar": "^2.1.0",
+    "bluebird": "^2.9.26"
```
tbranyen: This dependency isn't needed.  We use `nodegit-promise`.

test/tests/tree_entry.js
```diff
@@ -1,4 +1,5 @@
 var assert = require("assert");
+var promise = require("bluebird");
```
tbranyen: `var Promise = require('nodegit-promise')`

package.json
tomruggs: Thanks, not sure how I missed that.  Removed.

tbranyen:
Looks good, thanks!

PR #597, opened 2015-06-04
[test] Add tests to fully cover refs test

vanadium23:
I'm adding three test to fully cover lib/reference.js. Can you check that everything is alright?


tbranyen:
Looks good to me, thank you very much for writing these tests!

PR #598, opened 2015-06-05
Add test for RevParse object

vanadium23:
Another tests for good library. (:
I want to ask, if you would rather prefer small decoupled pull requests, or wait until there is more than N lines of code?


maxkorp:
Whatever works better for you is cool man. If you'd rather throw a big PR together, thats fine, or small ones work too.

johnhaley81:
Looks good @vanadium23 I'm going to merge this in.

PR #603, opened 2015-06-13
Make diff functions async

johnhaley81:
This is a pretty big change to how we use diffs. Right now they are all synchronous and can take a long time to calculate if the diff is large (>1k changes).

This PR will make the diff/hunk/lines all async so that at least it's not blocking when it's doing its calculations. This will cause an API break so it makes us have to do a minor version bump.

Also, the "can find similar files in a diff" test in `test/tests/diff.js` was incorrectly passing before. I'm skipping it right now until we can figure out why it's not passing.

@orderedlist I know this will affect you and what you wrote so if you want to chime in on this feel free :)

### DIFF
--- a/test/tests/diff.js
+++ b/test/tests/diff.js
@@ -91,58 +91,101 @@ describe("Diff", function() {
   });
 
   it("can walk a DiffList", function() {
-    var patch = this.diff[0].patches()[0];
+    return this.diff[0].patches()
+      .then(function(patches) {
+        var patch = patches[0];
 
-    assert.equal(patch.oldFile().path(), "README.md");
-    assert.equal(patch.newFile().path(), "README.md");
-    assert.equal(patch.size(), 1);
-    assert.ok(patch.isModified());
+        assert.equal(patch.oldFile().path(), "README.md");
+        assert.equal(patch.newFile().path(), "README.md");
+        assert.equal(patch.size(), 1);
+        assert.ok(patch.isModified());
 
-    var hunk = patch.hunks()[0];
-    assert.equal(hunk.size(), 5);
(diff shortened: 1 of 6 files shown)


johnhaley81:
Crap, I'll fix those falling tests when I get back home.

test/tests/diff.js
```diff
@@ -91,58 +91,101 @@ describe("Diff", function() {
...
-    assert(patches[2].isUntracked());
+    return this.workdirDiff.patches()
+      .then(function(patches) {
+        assert.equal(patches.length, 3);
+        assert(patches[2].isUntracked());
 
-    var oldFile = patches[2].delta.oldFile();
-    assert.equal(oldFile.path(), "wddiff.txt");
-    assert.equal(oldFile.size(), 0);
+        var oldFile = patches[2].delta.oldFile();
```
tbranyen: Are these tab characters?

johnhaley81: Yeah. I don't really know how that happened... Fixing.

johnhaley81: Hmm.. so after looking at the diff I'm not sure why those are highlighted. They aren't tabs in the raw file nor in my editor.

tbranyen: Okay awesome, weird!

orderedlist:
This is fine, though I *do* think we need to start pushing towards having both sync && async for some functions, similar to some of the core Node functionality. Overall though, I'm :+1:.

johnhaley81:
I agree with having async/sync options when we can. Anything that uses callbacks must be async or we'll deadlock. That's a different PR IMO. I'm going to go ahead and merge this in.

PR #614, opened 2015-06-22
if class has no functions, don't print it

heavyk:
like it says :)

cheers

### DIFF
--- a/generate/templates/templates/nodegit.js
+++ b/generate/templates/templates/nodegit.js
@@ -22,8 +22,10 @@ catch (ex) {
 {% each . as idef %}
   {% if idef.type != "enum" %}
 
-    var _{{ idef.jsClassName }}
-      = rawApi.{{idef.jsClassName}};
+    {% if idef.functions.length > 0 %}
+      var _{{ idef.jsClassName }}
+        = rawApi.{{ idef.jsClassName }};
+    {% endif %}
 
     {% each idef.functions as fn %}
       {% if fn.isAsync %}
@@ -98,7 +100,7 @@ require("./enums.js");
           _{{ idef.jsClassName }}.prototype;
 
         // Assign the function as the root
-        rawApi.{{idef.jsClassName}} =
+        rawApi.{{ idef.jsClassName }} =
           _{{ idef.jsClassName }}.{{ fn.jsFunctionName }};
 
       {% endif %}


tbranyen:
What's the motivation for this? Reducing the total exposed class count?

heavyk:
reduces the number of unused variables in the generated code

tbranyen:
@heavyk yeah, i'm just wondering if we have any classes that have enums attached that we shouldn't be removing.  i don't know off the top of my head.

heavyk:
well, the block right below it only loops over class and prototype methods, so there should be no problem.

there are still quite a few more unused variables too, but that would require extra checks, such as `_Indexer` `_Giterr`, `_Libgit2`, etc.

heavyk:
example to show further cleaning is possible
```js
var _Indexer = rawApi.Indexer; // <-- unused. can be removed
var _Libgit2 = rawApi.Libgit2; // <-- unused. can be removed
var _Merge = rawApi.Merge;

var _Merge_base = _Merge.base;
_Merge.base = promisify(_Merge_base);
```

tbranyen:
Well I'm good with this, I'll merge tomorrow barring nobody has objections.

heavyk:
so check it out, I got all the missing variables now. it was something I didn't know about ejs ... all I needed was `if idef.functions.length > 0` instead of `if idef.functions.length`

so, I guess in ejs truthy is if the prop exists (even if it's 0). I guess that makes sense if you want to tell someone they `you have 0 unread messages` instead of `you have  unread messages`

tbranyen:
@heavyk we aren't using EJS, it's a template engine called combyne.  https://github.com/tbranyen/combyne

heavyk:
oh, ok... so it's yours. nice
well, all that then with combyne :)

I don't have experience with template engines.
I have in fact never used even one.
closest is, I have to program angular at work. the filters looked similar

so yeah that is in fact why I made the mistake of using `if functions.length`, because I thought it'd be analogous to `if (idef.functions.length) { ... }` as if it were the contents of `ng-if="..."` but I didn't know how to put `ng-if` in the middle of c++ :)

cool, I'll have more of a look a bit later
cheers mate

heavyk:
final change added is a style change. as I was looking at the templates, I noticed that everywhere you have space around your curly braces, so I couldn't help myself... it was bothering me..

johnhaley81:
Restarted that failing test on Travis. Looks like it was a timeout. I'll merge after it passes.

Issue #599, opened 2015-06-08
Revwalk#getCommits() et al fail


alxarch:
Revwalk walk methods fail because they expect the instance to have a `repo` property.
[Here](https://github.com/alxarch/nodegit/commit/0fc6edbcd2f07b84fb96e4125f3854ce86196d45) is a patch to fix this

maxkorp:
This made it in at some point, so I'm going to close this

PR #627, opened 2015-06-27
Fix missing Revwalk#repo property

johnhaley81:
Most of the walk callbacks fail due to a missing `repo` property.


johnhaley81:
This should fix #599

johnhaley81:
@alxarch can you fix the linter errors? Then I think we can merge this in.

maxkorp:
I just went ahead and fixed this here: https://github.com/nodegit/nodegit/pull/669

PR #638, opened 2015-07-08
fixed string concatenation

num-lock:
i guess this was a typo, which results in unexpected output like `Foo Bar < foo.bar@example.com>` instead of `Foo Bar <foo.bar@example.com>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ open("tmp")
       var author = commit.author();
 
       // Display author information.
-      console.log("Author:\t" + author.name() + " <", author.email() + ">");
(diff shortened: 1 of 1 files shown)


tbranyen:
Our CI is busted ATM, not sure what happened.

johnhaley81:
@tbranyen looks like our HTTP test is dying. I think we should just skip this short-term to allow PR's to merge.

johnhaley81:
This is an issue with our CI and not the PR. I'm going to merge it anyways.

Thanks for the PR @num-lock!

PR #645, opened 2015-07-20
Add amend commit convenience function 

smith-kyle:
The convenience function extends the original commit amend to accepts a tree object or a treeOid.
Also included is a test for the new commit amend function.

When testing the new function we noticed that the add commit test did not reset the head to the previous commit upon completion of the test. We modified that test to do so. This caused the "can diff with a null tree" test to fail since it was expecting 85 patches and resetting the commit test brought the number of patches to 84. We modified the test to expect the correct number of patches.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -325,6 +325,26 @@
     },
     "commit": {
       "functions": {
+        "git_commit_ammend": {
+          "args": {
+            "id": {
+              "isReturn": true
+            },
+            "message_encoding": {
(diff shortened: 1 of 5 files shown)


johnhaley81:
@tbranyen do you know why it can't hit http://git.tbranyen.com/smart/site-content ?

tbranyen:
@johnhaley81 Not sure why it was failing before, but I just started my Node server and it seems to be cloning fine for me.

PR #648, opened 2015-07-21
Revert "C++ promisifyer"

johnhaley81:
This commit had an `eval` in it which is no bueno. Luckily reverting the commit was easy.


johnhaley81:
This isn't needed anymore because of #649

PR #651, opened 2015-07-23
Make `fetchAll` run in series to prevent deadlocking

johnhaley81:


lib/repository.js
```diff
@@ -648,14 +649,20 @@ Repository.prototype.fetchAll = function(
...
 
   return repo.getRemotes().then(function(remotes) {
-    var fetchPromises = [];
+    var fetchPromise = Promise.resolve();
 
     remotes.forEach(function(remote) {
-      fetchPromises.push(
-        repo.fetch(remote, remoteCallbacks, autoTag, pruneAfter, callback));
+      fetchPromise.then(function() {
+        repo.fetch(remote, remoteCallbacks, autoTag, pruneAfter, callback);
```
tbranyen: Isn't callback here supposed to be resolve? Or something else? I don't get how this could work

johnhaley81: Touché. So it shouldn't have that passed in at all. Ahh I didn't even return the promise so the chain isn't working as I thought it would. I'll get this fixed up.

johnhaley81:
@tbranyen so I don't think that it was the "Running in parallel" that was breaking `fetchAll` but rather that the promise chain was messed up originally and could put it in a weird state. This should fix that issue.

johnhaley81:
Still doing more testing to make sure that this is working for a lot of remotes. What I was seeing was intermittent.

johnhaley81:
This is looking good now. After a green-light on CI I'm going to merge this.

johnhaley81:
So after another hour or so of monitoring a `fetchAll` loop that I have I've gotten it to stall again. So I switched it back to synchronous and I'm going to monitor that now for a bit.

Very strange.

I think it might be related to the stuff we saw with the `Clone.clone` tests and things needing to be freed.

johnhaley81:
So far 2 hours no deadlocking. I think that fetching DOES need to be synchronous. Any objections to this change?

johnhaley81:
Ran overnight with no deadlocks. Still doesn't makes me a little nervous that I can't exactly pin point what's deadlocking but this does seem to fix it.

Issue #655, opened 2015-07-29
Repository.createCommitOnHead ignores `filesToAdd` argument


freder:
hi,

it looks like `Repository.createCommitOnHead` ignores the first argument `filesToAdd`. it is not using it at all, and instead [adds all the files!](https://github.com/nodegit/nodegit/blob/master/lib/repository.js#L496)

and that is why this function to create empty commits doesn't work as expected when there are (unstaged) files with changes:

``` javascript
function createEmptyCommit(repo, message, sig) {
    var files = []; // <--
    sig = sig || createSignature();
    var author = sig;
    var committer = sig;

    return unstageAll(repo) // make sure commit is in fact empty. (does a mixed reset)
        .then(function() {
            return repo.createCommitOnHead(
                files,
                author,
                committer,
                message
            );
        })
        .catch(function(err) {
            handleError(err);
        });
}
```

freder:
using this for now

```javascript
function createCommitOnHead(filesToAdd, author, committer, message) {
	var index, repo;
	return getRepo()
		.then(function(repo_) {
			repo = repo_;
			return repo.openIndex();
		})
		.then(function(index_) {
			index = index_;
			index.read(1);
			if (!filesToAdd) { filesToAdd = []; }
			var result = filesToAdd.map(function(item) {
				return index.addByPath(item);
			});
			return Q.all(result);
		})
		.then(function() {
			index.write();
			return index.writeTree();
		}).then(function(treeOid) {
			return repo.getHeadCommit()
				.then(function(parent) {
					return repo.createCommit(
						'HEAD',
						author,
						committer,
						message,
						treeOid,
						[parent]
					);
				});
		}).
		catch(handleError);
};
```

tbranyen:
Can you PR that?

rcjsuen:
Can this be closed? #656 was closed but the corresponding #670 has been merged into `master`.

PR #656, opened 2015-08-03
only add filesToAdd — not all

freder:
#655

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -491,12 +491,14 @@ Repository.prototype.createCommitOnHead = function(
   var index;
   var repo = this;
 
-  repo.openIndex()
+  return repo.openIndex()
     .then(function(index_) {
       index = index_;
       index.read(1);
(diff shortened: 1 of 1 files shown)


lib/repository.js
```diff
@@ -486,32 +486,33 @@ Repository.prototype.createCommitOnHead = function(
...
-        "HEAD",
-        author,
-        committer,
-        message,
-        treeOid,
-        [parent],
-        callback);
-    });
-  }, callback);
+  repo.openIndex()
```
johnhaley81: This still needs to return the promise so we can get the value.

freder:
[updated](https://github.com/nodegit/nodegit/commit/a49db638b5d32ad0d63269fe5a2b33edbb390c2c)

maxkorp:
Fixed the lint errors here: https://github.com/nodegit/nodegit/pull/670

Thanks for making this PR man.

PR #659, opened 2015-08-04
Fix electron build

johnhaley81:


johnhaley81:
Closed in lieu of #660

PR #663, opened 2015-08-06
Adding convenience methods for setUrl and setPushurl methods on the Remote object

cbargren:
After talking to Jose and Tyler, likely what happened is libgit2 changed the signature of git_remote_set_url to take in remote as a `char *` instead of a `git_remote *`. This caused the generator to not recognize setUrl as an instance method. They're in the NodeGit documentation and they're useful methods so adding them to the Remote prototype as convenience methods seems reasonable.

### DIFF
--- a/lib/remote.js
+++ b/lib/remote.js
@@ -124,4 +124,24 @@ Remote.prototype.push = function(refSpecs, opts) {
   return push.call(this, refSpecs, opts);
 };
 
+/**
+ * Set the remote's url in the configuration
+ *
+ * @param {String} url The url to set
+ * @return {Number} error code
+ */
+Remote.prototype.setUrl = function(url) {
+  return NodeGit.Remote.setUrl(this.repo, this.name(), url);
+};
+
+/**
+ * Set the remote's url for pushing in the configuration
+ *
+ * @param {String} url The url to set
+ * @return {Number} error code
+ */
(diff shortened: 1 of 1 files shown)


cbargren:
I did just realize that it wouldn't actually update the url of the remote you pass in. Might need to do something about that.

johnhaley81:
You mean @jdgarcia and @implausible?

johnhaley81:
So the changes in libgit2's API for remotes will now make is so that this won't change the values in memory for that remote. We can't really do much on our end to fix this.

So I kind of don't like these methods being on the remote anymore though because it does imply that you are setting the remote instance's values to what you are passing in. Having them live on `NodeGit.Remote` makes it seem at least that you're not modifying the instance you have (which is correct).

cbargren:
@johnhaley81 Yeah, I'm no longer convinced that this is quite a great idea. The convenience methods were nice to have and it might be somewhat strange at first glance that they're not there, but it would be even stranger that `setUrl` doesn't actually set the url of the object you're calling the method on. Unless @jdgarcia or @implausible think otherwise, I'm okay with leaving these out.

johnhaley81:
Yeah, I'm thinking that they should stay on `Remote.setUrl` and not the prototype. Closing this out. Thanks though!

PR #674, opened 2015-08-12
Update supported Node JS versions

ChristianMurphy:
Adds io.js 3 to the Travis CI environment test matrix

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,8 +26,8 @@ branches:
 
 before_install:
   - if [ $TRAVIS_OS_NAME == "osx" ]; then
-      git clone https://github.com/creationix/nvm.git ./.nvm
-      source ./.nvm/nvm.sh
+      git clone https://github.com/creationix/nvm.git ./.nvm;
+      source ./.nvm/nvm.sh;
     fi
   - nvm install $NODE_VERSION
   - nvm use $NODE_VERSION
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,16 @@ os:
   - linux
   - osx
 
-node_js:
-  - '0.10'
-  - '0.12'
-  - iojs-2
-  - iojs-3
+env:
+  - NODE_VERSION="0.10"
+  - NODE_VERSION="0.12"
+  - NODE_VERSION="iojs-2"
+  - NODE_VERSION="iojs-3"
 
 matrix:
   fast_finish: true
   allow_failures:
-    - node_js: iojs-3
+    - env: NODE_VERSION="iojs-3"
 
 git:
   depth: 1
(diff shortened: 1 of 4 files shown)


ChristianMurphy:
[Errors thrown](https://travis-ci.org/nodegit/nodegit/jobs/75297165#L752) appear to be related to the [V8 update in io.js 3](https://github.com/nodejs/node/wiki/Breaking-Changes#v8-upgrade)

johnhaley81:
I think this will require a bump to the latest [nan](https://github.com/nodejs/nan/releases/tag/v2.0.5)

ChristianMurphy:
bumping nan to 2.0.5 allows support for io.js 3, but currently appears to break many of the openssl and  libgit2 bindings. :frowning:

tbranyen:
I've said it before, and unfortunately again.  io.js is a great moving target, but a nightmare to support with native code.  We'll continuously support Node, but we'll need an io.js advocate who will maintain parity if we want to support latest.  I don't think it's fair to ask our devs who have little time as-is to support something that breaks as often.

ChristianMurphy:
@tbranyen fair enough.
I removed the io.js 3 tests, updated engines to give a warning if version 3+ is used and added a note in the `README` on support.
Would it be safe to say that [Node JS Long Term Support](https://github.com/nodejs/LTS) releases will be supported by NodeGit?

tbranyen:
@ChristianMurphy "It should be stated that the abstraction layer (currently NAN) should support all current LTS releases. " If that can hold true, then absolutely!

tbranyen:
@ChristianMurphy according to this: https://medium.com/node-js-javascript/4-0-is-the-new-1-0-386597a3436d it's in our best interest to support iojs v3

ChristianMurphy:
Cool, would you like me to re-add the `io.js 3` tests? (probably as allow failure for now)

tbranyen:
@ChristianMurphy so it's only failing in Windows then? surprising

tbranyen:
Oh derp, allowed failure

ChristianMurphy:
Yep, and the build failure on Windows appears to be AppVeyor flaking out rather than an actual build failure.
Thoughts on the changes made in the PR? :thought_balloon:

tbranyen:
@ChristianMurphy can you bump us to nan 2.0 in here too?

ChristianMurphy:
@tbranyen updated :+1:

johnhaley81:
Well that broke everything :)

tbranyen:
Merge it!

dukex:
Why not use 'node_js' config? 
http://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Choosing-Node-versions-to-test-against

ChristianMurphy:
OS X build currently do not support the `node_js` config.

Ref https://github.com/travis-ci/travis-ci/issues/2311

dukex:
:/  my  bad, thank you @ChristianMurphy

ChristianMurphy:
@dukex no worries :smile: 
It threw me for a loop when I tried it in e929f76 and everything blew up :frowning: 
Hopefully Travis will eventually add Node JS for OS X to their [roadmap](http://next.travis-ci.com/)

johnhaley81:
Trying to get nan v2 working in #702

PR #684, opened 2015-08-19
Implemented private remote tests

implausible:
Added test to check if nodegit can fetch from a private repo with valid creds

Added test to check if nodegit cannot fetch from private repo with invalid creds


tbranyen:
Wow thanks!

test/tests/remote.js
```diff
@@ -163,6 +164,63 @@ describe("Remote", function() {
...
+      }
+    };
+
+    return remote.fetch(null, fetchOptions, "Fetch from private")
+      .catch(function() {
+        assert.fail("Unable to fetch from private repository");
+      });
+  });
+
+  it("can reject fetching from private repository without valid " +
```
tbranyen: One minor thing here.  Is it necessary to concat here?

implausible: Changed so it that function () { is on the next line, it's necessary to split up the it("string" and the function definition. What's preferable? it("", \nfunction\n{ or it("", \nfunction() { ?

tbranyen: If you keep it on the same line does it fail JSHint or something?

implausible: yup

tbranyen: Okay, oh well ha.  Thanks!

PR #685, opened 2015-08-20
Bump `promisify-node` to 0.2.1

johnhaley81:
This makes our promisified functions use the latest `nodegit-promise` library (3.0.2)

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "node-pre-gyp": "^0.6.5",
     "nodegit-promise": "^3.0.1",
     "npm": "^2.9.0",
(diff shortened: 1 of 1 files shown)


ephaee:
Is there a reason nodegit does not use native v8 Promises (I believe that was introduced in node.js 0.10)?

maxkorp:
A) node 10 required harmony flags for promises
B) we require synchronous polling of the promise objects so that we can check their state to do some thread sleeping in C++, which native promises don't provide.

ephaee:
thanks

maxkorp:
No problemo man.

PR #687, opened 2015-08-24
Drop support for  node 0.10

maxkorp:
This makes it way easier to deal with the electron (really, node-pre-gyp) madness, and everyone is cool with it.

 - Updates the engines field in package.json
 - Removes all of the references to (and handling for) legacy http_parser
 - Removes 0.10 from test matrices

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 env:
   matrix:
-    - export NODE_VERSION="0.10"
     - export NODE_VERSION="0.12"
     - export NODE_VERSION="iojs-v2.5.0"
 matrix:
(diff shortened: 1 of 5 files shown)


maxkorp:
Worth noting that iojs2 has always failed on appveyor, and is due to some weird npm config error (that he even gets on a brand new box). Working on fixing that elsewhere, but iojs2 builds and tests fine locally as before (as this shouldn't affect it).

johnhaley81:
I'm :+1: on this.

maxkorp:
KK, I know @tbranyen said he was cool with it in gitter, so in it goes!

PR #691, opened 2015-08-30
Mark the output param of treebuilder.write. Its unusable otherwise be…

tannewt:
…cause one cannot instantiate an Oid to pass in.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2059,6 +2059,13 @@
       "functions": {
         "git_treebuilder_filter": {
           "ignore": true
+        },
+        "git_treebuilder_write": {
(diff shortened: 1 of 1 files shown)


tbranyen:
Awesome thanks, is this something you'd feel comfortable writing a unit test for?

tannewt:
I can. Do have an example I can base it off of?

tbranyen:
Doesn't look like we have a treebuilder file yet, create one and base off of:

https://github.com/nodegit/nodegit/blob/master/test/tests/tree_entry.js

tbranyen:
Would rather have this in than out.  We can revisit a test later.

PR #693, opened 2015-08-31
Link to building from source page

petecoop:
Building from source instructions aren't on the readme!


tbranyen:
Good call, thanks!

PR #694, opened 2015-08-31
Allow update_ref to be null in Commit.amend

srajko:
NodeGit's http://www.nodegit.org/api/commit/#amend chokes if you give it a null update_ref value, but according to https://libgit2.github.com/libgit2/#HEAD/group/commit/git_commit_amend NULL is acceptable.

This PR changes update_ref to an optional parameter and adds a test.


tbranyen:
A model PR, thank you so much!

PR #698, opened 2015-09-04
Don't include 'No newline at end of file' when staging lines and hunks.

kenprice:


tbranyen:
@kenprice looks like JSHint issues:

```
lib/repository.js: line 1194, col 98, Line is too long.
lib/repository.js: line 1194, col 89, Strings must use doublequote.
```

PR #699, opened 2015-09-04
Stage line eof bug

kenprice:
Fixes bug that involves staging lines:
- User can stage '\ no newline at end of file'. Shouldn't happen!
- Last character in a diff gets cut off if it's the last line in file.
- Staging the last line with no \n will magically add a newline. Shouldn't happen.


johnhaley81:
Those edge cases... Thanks @kenprice!

PR #703, opened 2015-09-09
Stage line eof bug part 2

kenprice:
This one's better. Accounts for a case I haven't considered previously.
Basically added
```
      if (hunkLine.raw.content().indexOf("\\ No newline at end of file") != -1) {
        return false;
      }
```
in `Repo.prototype.stageLines`'s `processSelectedLine`. This replaces a similar check for `\\ No newline at end of file` I originally had in `Repo.prototype.stageLines`.


johnhaley81:
Looks good. We'll at least have test for this now going forward. Thanks!

PR #707, opened 2015-09-14
Node 4: Use embedded OpenSSL headers first

saper:
This a workaround for the compilation
error with Node 4:

../vendor/openssl/openssl/ssl/s3_clnt.c:3262:60: error: use of undeclared identifier 'EVP_PKT_EXP'

Node 4.0.0 includes openssl 1.0.2d, which
no longer defines EVP_PKT_EXP and friends.

node-gyp puts header directories supplied
by node first, therefore OpenSSL include
files shipped by node are used, and
not the ones from the embedded OpenSSL
source.

### DIFF
--- a/vendor/libgit2.gyp
+++ b/vendor/libgit2.gyp
@@ -512,7 +512,7 @@
         "_REENTRANT",
         "NO_WINDOWS_BRAINDEATH",
       ],
-      "include_dirs": [
+      "include_dirs+": [
         ".",
         "openssl/openssl",
         "openssl/openssl/crypto",


tbranyen:
Wow, brilliant, thank you @saper!

saper:
On Mon, 14 Sep 2015, Tim Branyen wrote:

> Wow, brilliant, thank you @saper!

Why can't we use OpenSSL that ships with node?

tbranyen:
I'm not sure, if you can figure out how to make it work, we'll happily accept a PR.

saper:
> I'm not sure, if you can figure out how to make it work, we'll happily accept a PR.

ah, ok, I wasn't sure if this was given any consideration and there were some issues
which made us not do it.

tbranyen:
@saper I vaguely remember trying to get it work ages ago, I think @TooTallNate or someone else had info on linking to it wrt node-gyp.  I think we gave up due to Windows issues and decided to build our own.

TooTallNate:
Still an issue unfortunately: https://github.com/nodejs/node-v0.x-archive/issues/4051

saper:
Thanks @ToolTallNate. So what we need is the intelligent way, cross-platform way to create list of the APIs exported (to generate a linker export map)?

saper:
Thanks @TooTallNate. So what we need is the intelligent way, cross-platform way to create list of the APIs exported (to generate a linker export map)?

PR #708, opened 2015-09-14
Fix error: invalid operands to binary expression ('size_t' (aka 'unsigned long') and 'nullptr_t')

saper:
When compiling with clang we are getting:

    ../src/checkout_options.cc:536:39:
       error: invalid operands to binary expression
       ('size_t' (aka 'unsigned long') and 'nullptr_t')

This commit improves the generated code.


tbranyen:
Thank you!

PR #715, opened 2015-09-15
Drop recursive npm install

saper:
By the time "install" lifecycle script
is invoked npm already has installed
all the dependencies.

### DIFF
--- a/lifecycleScripts/install.js
+++ b/lifecycleScripts/install.js
@@ -97,21 +97,19 @@ function build() {
     target = "--target=" + nwVersion;
   }
 
-  return exec("npm install --ignore-scripts")
-    .then(function() {
-      builder = path.resolve(".", "node_modules", ".bin", builder);
-      builder = builder.replace(/\s/g, "\\$&");
-      var cmd = [prefix, builder, "rebuild", target, debug, distUrl]
-        .join(" ").trim();
+  builder = path.resolve(".", "node_modules", ".bin", builder);
+  builder = builder.replace(/\s/g, "\\$&");
+  var cmd = [prefix, builder, "rebuild", target, debug, distUrl]
+    .join(" ").trim();
 
-      return exec(cmd, opts);
-    })
+  return exec(cmd, opts)
     .then(function() {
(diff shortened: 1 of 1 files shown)


saper:
It seems to fix issues with iojs 2.x build in AppVeyor.

maxkorp:
Shoot, are we sure this works for git-referenced npm installs? Like in your package.json
```json
"dependencies": {
  "nodegit": "nodegit/nodegit"
} 
```
It (at least) used to be the case that in that scenario, it wouldn't install devdependencies, only dependencies. But we needed the devdependencies installed to do the generation. Has that changed with newer versions of NPM?

saper:
> Shoot, are we sure this works for git-referenced npm installs? Like in your package.json
> ```json
> "dependencies": {
>   "nodegit": "nodegit/nodegit"
> } 
> ```

Haven't tested that...

> It (at least) used to be the case that in that scenario, it wouldn't
> install devdependencies, only dependencies. But we needed the
> devdependencies installed to do the generation. Has that changed with
> newer versions of NPM?

In that case we just need to move stuff from dev to normal (unfortunately).
Unless we make it work...

maxkorp:
Ah yeah, thats why we do that. 
Fortunately I've managed to chop off a lot of the dev dependencies at least. This is actually an issue I've been talking with the NPM folks for a long time now (the ability to run some scripts before install, but after deps are installed). 

Actually, could we include those deps as bundled deps? IIRC in that case npm will go fetch those first. 
If not, we have to move a few packages, but more can be deleted

- remove npm from deps (requires deleting lifecyclescripts/clean, but thats been dead for a while)
- remove `tar` and `request` from dev-deps
- move `combyne`, `js-beautify`, `lodash`, `nan`, `node-gyp`, and `nw-gyp` to deps

Additionally, with some love (that I keep getting started on then pulled from), we could get rid of the dependency on `nw-gyp` and `which-native-nodish`. With a PR to node-pre-gyp, we may be able to ditch the dep on `node-gyp` as well, if they bundle it with themselves.

PR #722, opened 2015-09-16
Stage last hunk? No empty diffs!

kenprice:
Upon staging a hunk/line, stage the whole file if:
* It is the last hunk/line, ie the result of 
```Diff.indexToWorkdir(repo, index, {
          flags:
            NodeGit.Diff.OPTION.SHOW_UNTRACKED_CONTENT |
            NodeGit.Diff.OPTION.RECURSE_UNTRACKED_DIRS
        })```
  will be a diff containing a single patch with _no hunks_.
* The file does not have filemode (permission, type) changes.

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -1242,6 +1242,39 @@ Repository.prototype.stageLines =
           NodeGit.Diff.OPTION.RECURSE_UNTRACKED_DIRS
       });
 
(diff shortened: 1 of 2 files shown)


johnhaley81:
Great find @kenprice. Thanks for the PR.

PR #724, opened 2015-09-16
Fix import path in example

danawoodman:
The example is importing relative when it should import the named module.


tbranyen:
Thanks!

danawoodman:
:+1: my pleasure!

PR #725, opened 2015-09-17
Let libssh2 configure generate its output

saper:


samkelleher:
@saper What does the spawn vs exec do here? Is a new process starting before a previous one has finished or something?

maxkorp:
It's allowing him to pipe stdio and stderr from the child process up to the main process.

Issue #712, opened 2015-09-15
npm test multilates user's .gitconfig


saper:
I just realized that I am push my commits as `undefined` to GitHub aftery trying `nodegit` yesterday.

My `.gitconfig` got mutilated by `npm test` - via `git config --global --unset ...`.

This is not acceptable. One should set `HOME` and/or `userprofile` to some fake user home directory before doing this!

tbranyen:
Wow that's incredibly surprising as this project has existed for literally years with contributors and never had this issue brought up.  Can you pinpoint exactly where this happens?

saper:
I just did `git grep 'git config'` and saw those horrible things in tests.

tbranyen:
Urgh, yeah, that's a bit naive we shouldn't be modifying the global configuration for any reason.

tbranyen:
I like your idea about a testing home directory setup, we should definitely commit something like that.

johnhaley81:
Fixed in #727

PR #727, opened 2015-09-18
Set HOME environment variable for tests.

kenprice:
Addresses https://github.com/nodegit/nodegit/issues/712
by setting HOME to some dummy directory.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -15,9 +15,11 @@ var args = cov.concat([
   "--expose-gc"
 ]);
 
-var local = path.join.bind(path, __dirname);
-var dummyPath = local("home");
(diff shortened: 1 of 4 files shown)


test/index.js
```diff
@@ -15,4 +15,8 @@ var args = cov.concat([
   "--expose-gc"
 ]);
 
+var local = path.join.bind(path, __dirname);
+var dummyPath = local("home");
+process.env.HOME = dummyPath;
```
saper: We should probably also add `process.env.USERPROFILE = dummyPath;` for Windows

johnhaley81:
Thanks for getting that for us @kenprice!

PR #729, opened 2015-09-20
Set HOME properly

saper:


tbranyen:
Nice! This is similar to https://github.com/nodegit/nodegit/pull/727 right? If so, should we add the same Git config functionality

saper:
Not sure what you mean... This is to change location of files fetched by node-gyp. I've never worked with electron but  
I cnc image it is necessary. This has no relation to the git configuration.

PR #739, opened 2015-10-06
Fix problems with async code

implausible:
In asynchronous libgit calls with callbacks, when the callback fails or returns an error, the asynch isn't considered a failure. It also loops forever, even after the promise has rejected. The promise should then instead reject, return an error to the appropriate caller, and return that error to the javascript layer on nodegit as a rejection.

Now when a promise rejects in a callback, this should call the function calling the callback to reject as well.


johnhaley81:
@implausible can you make another PR to update libgit2? This is way to huge to figure out what's going on.

implausible:
See https://github.com/nodegit/nodegit/pull/743 for libgit2 bump

generate/templates/partials/field_accessors.cc
```diff
@@ -258,7 +261,7 @@
         else {
           // promise was rejected
           baton->result = {{ field.return.error }};
-          baton->done = false;
+          baton->done = true;
```
johnhaley81: I don't remember why we did this originally but this change makes a lot more sense now. :+1:

johnhaley81:
I'm good on this. Thanks @implausible !

PR #740, opened 2015-10-07
Added diff functions with options parameter.

ayresa:
Allows options for diffs so that we can do things like exclude whitespace for example.


johnhaley81:
@tbranyen do you know why coverage is dying on linux?

tbranyen:
Nope, I'll wait till the rebuild happens and check.

tbranyen:
Looks fine :-p

johnhaley81:
Nothing to see here :)

PR #743, opened 2015-10-08
Bump libgit2

implausible:
There's a change made to the file vendor/libgit2/src/xdiff/xdiff.h:
#include "util.h" -> #include "../util.h"

This causes mac osx to fail merges on osx/linux because it causes git__malloc to be undefined.


implausible:
Test failed because of timeout on travis. Someone please restart it!

johnhaley81:
Thanks for also getting rid of the file mode changes as well.

PR #744, opened 2015-10-08
Add instructions when using on Travis CI

Coaxial:
If one is using Travis CI to build a project that depends on nodegit, the build will fail because it can't compile.

These steps will keep this from happening and will let nodegit compile happily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,3 +201,24 @@ the change to nan v1.4.3 doesn't cause any problems.
 
 Currently, support for nw.js is limited, although we intend to support it better
 in the future.
+
+## Travis CI support
+
+If you are using nodegit within a project that is built with Travis CI and you are getting failures when compiling about a GLIBC version mismatch, add the following to your `.travis.yml` file:
+
+```yml
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+    packages:
+      - g++-4.9
+      - gcc-4.9
+      - libc6-dev
+
+before_install:
+  - export CC=/usr/bin/gcc-4.9;
+  - export CXX=/usr/bin/g++-4.9;
+```
+
+It will install the required dependencies and make sure the right versions are used when compiling.


tbranyen:
This is useful information if the build does fail, although with latest Node 0.12 or 4.1 it should download the prebuilt binary that we publish.

Coaxial:
It doesn't though, see the log for the latest build I did:

```
395.37s$ npm install 
> nodegit@0.5.0 install /home/travis/build/Coaxial/coverall/node_modules/nodegit
> node lifecycleScripts/install
[nodegit] Fetching binary from S3.
[nodegit] Failed to install prebuilt binary, building manually.
[nodegit] Regenerating and configuring code
```

I'm not sure why this happens but it makes build take forever because it has to recompile every time.

maxkorp:
Can you give us your platform (nix/osx?), arch, and node/npm versions?

Coaxial:
Sure, this occurs on Travis CI when testing against node 0.10, 0.11 and 0.12. I believe they are using Ubuntu 12.04 in Docker containers.

On 4.0 and 4.1, it fails with the following, no matter what I tried:
```
Message:
    /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.17' not found (required by /home/travis/build/Coaxial/coverall/node_modules/nodegit/build/Release/nodegit.node)
Stack:
Error: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.17' not found (required by /home/travis/build/Coaxial/coverall/node_modules/nodegit/build/Release/nodegit.node)
    at Error (native)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/travis/build/Coaxial/coverall/node_modules/nodegit/lib/nodegit.js:8:12)
    at Module._compile (module.js:434:26)
    at Object.Module._extensions..js (module.js:452:10)
    at Object.Module._extensions.(anonymous function) [as .js] (/home/travis/build/Coaxial/coverall/node_modules/gulp-istanbul/node_modules/istanbul/lib/hook.js:109:37)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at Object.<anonymous> (/home/travis/build/Coaxial/coverall/lib/archive.js:9:269)
    at Module._compile (module.js:434:26)
    at Object.Module._extensions.(anonymous function) [as .js] (/home/travis/build/Coaxial/coverall/node_modules/gulp-istanbul/node_modules/istanbul/lib/hook.js:107:24)
```

Coaxial:
By the way, why not use bluebird's promisify and bluebird's promises rather than `node-promisify`?

tbranyen:
@Coaxial someone tried to use Bluebird in the past, but it broke wayyyy too much stuff. I don't remember the specifics, but you can crawl through the history log for the revert.

We do not support 0.10 for precompiled builds, which means it will compile in those cases. Node 0.12 and Node 4.1 on Travis should not be compiling, can you verify this and/or show us where that's not the case?

johnhaley81:
This might no longer be needed after #761

contolini:
We're having a similar issue. Travis fails to install the precompiled binary and [eventually dies](https://travis-ci.org/contolini/CFPBot/builds/88645496#L1781-L1782) during build on v0.12.

When installing nodegit from master (`npm i --save nodegit/nodegit#master`), which includes @johnhaley81's #761, it still fails to install the precompiled binary but [successfully builds](https://travis-ci.org/contolini/CFPBot/builds/88652166#L666-L667) and everything is A-OK.

johnhaley81:
@contolini installing nodegit from the `master` branch won't download pre-built binaries since there are none published for master. We're looking to release again soon and when we do we'll have the pre-built binaries up for downloading during install.

maxkorp:
I think we finally git rid of the glibc requirement correct? If so this can just be closed
CC @johnhaley81

johnhaley81:
Yeah, we can close this one down. You don't need that anymore to build. If that changes for any reason we can open this back up. Thanks @Coaxial

runspired:
@johnhaley81 this is still an issue

johnhaley81:
Reopening due to #853

tbranyen:
I've since added instructions that are easier and don't require rebuilding the module (prebuilt works fine), so closing this.

PR #745, opened 2015-10-09
Modified "Diff can diff with a null tree" test.

ayresa:
Number of patches returned is 84 or 85 depending on something unknown at this time.  Hopefully we can eventually resolve the root cause of the difference.

### DIFF
--- a/test/tests/diff.js
+++ b/test/tests/diff.js
@@ -229,7 +229,7 @@ describe("Diff", function() {
         return diff.patches();
       })
       .then(function(patches) {
-        assert.equal(patches.length, 84);
+        assert.ok(patches.length === 84 || patches.length === 85);
       });
   });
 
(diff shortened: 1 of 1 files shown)


johnhaley81:
Something is clearly polluting the test environment as you run tests since this specific test can pass/fail if other tests are run or skipped. We should track this down but for right now we can use this as a work-around.

test/tests/diff.js
```diff
@@ -229,7 +229,7 @@ describe("Diff", function() {
         return diff.patches();
       })
       .then(function(patches) {
-        assert.equal(patches.length, 84);
+        assert.ok(patches.length === 84 || patches.length === 85);
```
johnhaley81: Can you add a comment here? Future us is going to look at this and be like "what the hell is happening?? Lets fix this". Add a link to #746 in the comment as well please.

PR #747, opened 2015-10-12
Fixed fetch call.

aral:
Fetch requires the credentials function to be wrapped in the callbacks object (otherwise the promise succeeds even though it has failed silently).


tbranyen:
:+1:

PR #750, opened 2015-10-12
Fix callback rejections

implausible:
callbacks don't return their rejection reason to the calling function, but only return a limited error.

### DIFF
--- a/test/tests/remote.js
+++ b/test/tests/remote.js
@@ -245,7 +245,7 @@ describe("Remote", function() {
     });
   });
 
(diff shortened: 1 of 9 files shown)


test/tests/remote.js
```diff
@@ -3,7 +3,7 @@ var path = require("path");
 var Promise = require("nodegit-promise");
 var local = path.join.bind(path, __dirname);
 
-describe("Remote", function() {
+describe.only("Remote", function() {
```
johnhaley81: =/

Issue #742, opened 2015-10-08
Does `fetch` work on bare repositories?


jimCresswell:
I'm cloning into a bare repo. When I want to fetch new branches and commits I use 
``` javscript
git.Repository.openBare(localPath)
    .then(function (repository) {
      return repository.fetch(repository.getRemote('origin'), {
        callback: ignoreSSLErrors,
        downloadTags: 1
      });
    })
```
This doesn't seem to have any effect on the state of the repository. Is there another step I need to see new branches or new commits on existing branches?

cc @sponte

sponte:
@jimCresswell - either way we should add following fetch options when fetching:

downloadTags and prune.

http://www.nodegit.org/api/fetch_options/

On 08/10/2015, Jim Cresswell <notifications@github.com> wrote:
> I'm cloning into a bare repo. When I want to fetch new branches and commits
> I use
> ``` javscript
> git.Repository.openBare(localPath)
>     .then(function (repository) {
>       return repository.fetch(repository.getRemote('origin'), {
>         callback: ignoreSSLErrors,
>         downloadTags: 1
>       });
>     })
> ```
> This doesn't seem to have any effect on the state of the repository. Is
> there another step I need to see new branches or new commits on existing
> branches?
>
> cc @sponte
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/nodegit/nodegit/issues/742


-- 
Kind regards,
Stanislaw Wozniak

aral:
I am seeing this same behaviour on a non-bare repository. No error, promise gets fulfilled but no fetch has occurred. Still assuming that I’m doing something wrong as fetch is most likely not generally broken.

jimCresswell:
In case it is relevant: I'm always cloning and fetching over https with no credentials, the repo test for fetch seems to use SSH.

aral:
The problem I was having was that the fetch example had a bug https://github.com/nodegit/nodegit/blob/master/examples/fetch.js#L7 where it was not passing a correctly-formed options object (pull request: https://github.com/nodegit/nodegit/pull/747).

[...]

PR #755, opened 2015-10-15
In repository.fetch don't swallow errors thrown by remote.fetch .

jimCresswell:
In repository.fetch don't swallow errors thrown by remote.fetch, catch them, pass them to the callback and then rethrow them.

There is a failing test "can reject fetching from private repository without valid credentials" for remote.js but that is not caused by this work.

Helps with https://github.com/nodegit/nodegit/issues/742

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -640,9 +640,9 @@ Repository.prototype.fetch = function(
 {
   var repo = this;
 
-  function finallyFn() {
+  function finallyFn(error) {
     if (typeof callback === "function") {
-      callback();
+      callback(error);
     }
   }
 
@@ -653,7 +653,11 @@ Repository.prototype.fetch = function(
         return remote.disconnect();
       });
     })
-    .then(finallyFn, finallyFn);
+    .then(finallyFn)
+    .catch(function(error) {
+      finallyFn(error);
+      throw error;
+    });
 };
 
 /**


tbranyen:
@jimCresswell I've been seeing that failure as well, but haven't had time to look into why it fails locally, but not in CI

jimCresswell:
@tbranyen something to do with the username and ssh key data injected before the Travis run? Even if they are invalid maybe they need to exist in that invalid state for the test to pass?

There appears to be a build failure for Mac using Node 0.12, the build worked with Mac and Node 4 so maybe it was a temporary network issue, is there a 'rebuild' button on your side?

lib/repository.js
```diff
@@ -653,7 +653,11 @@ Repository.prototype.fetch = function(
         return remote.disconnect();
       });
     })
-    .then(finallyFn, finallyFn);
+    .then(finallyFn)
```
johnhaley81: We could simplify this further by using `.finally` in [then/promise](https://github.com/then/promise/blob/master/src/finally.js#L6-L16)

jimCresswell: Is .finally part of the Node in Promise API? Also, in my case I'd like to catch errors outside of this scope in my own code, would that be affected?

jimCresswell:
Given that the failing test is (very probably) not related to this change are there any objections to merging it?

johnhaley81:
@jimCresswell I think this is pretty good to merge. Can you just rebase this on top of master and then we can merge it.

jimCresswell:
@johnhaley81 sure, rebase done.

maxkorp:
Appveyor error was a network flub, restarted

jimCresswell:
Green! :green_apple:

PR #763, opened 2015-10-21
Define GIT_SSH_MEMORY_CREDENTIALS for libgit2

### LINKED ISSUE #760: Prebuilt binary libgit2 not built with GIT_SSH_MEMORY_CREDENTIALS defined
Hi! When I call Cred.sshKeyMemoryNew, I get this error message:

```
Error: This version of libgit2 was not built with ssh memory credentials.
    at Error (native)
```

"npm install" installed prebuilt binaries for nodegit.

I'm running Node 4.0.0 on x86_64 Linux. THis is nodegit 0.5.0.

I believe this means libgit2 [should be built with GIT_SSH_MEMORY_CREDENTIALS defined](https://github.com/libgit2/libgit2/blob/master/src/transports/cred.c#L203).

I'm filing this after talking to @johnhaley81 in gitter.

daxelrod:
Compile libgit2 with GIT_SSH_MEMORY_CREDENTIALS defined to allow
sshKeyMemoryNew to work.

Fixes #760

### DIFF
--- a/test/tests/cred.js
+++ b/test/tests/cred.js
@@ -1,5 +1,6 @@
 var assert = require("assert");
 var path = require("path");
+var fs = require("fs");
 var local = path.join.bind(path, __dirname);
 
 describe("Cred", function() {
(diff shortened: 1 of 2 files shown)


daxelrod:
I guess I'll have to do more digging to find out why Windows doesn't like this define. Any insight would be appreciated.

johnhaley81:
It looks like libssh2 needs an update. That function isn't in the current implementation in `vendor/libssh2`

maxkorp:
Bumped to the latest libssh2 in #830, once that lands (if lol), this can be rebased to see if that fixes it.

maxkorp:
https://github.com/nodegit/nodegit/pull/1071 is merged, let's do this up guyz

maxkorp:
Fixed here: https://github.com/nodegit/nodegit/pull/949

PR #769, opened 2015-10-29
Add support for patch method git_patch_line_stats + test for it

jfremy:
I've just added support for the git_patch_line_stats method.
The only caveat is that the returned values are size_t being cast to a js number.
This is convenient to quickly gather stats about the patches without having to go down to the line level.

Added a simple test for the function as well

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -1360,7 +1360,20 @@
           }
         },
         "git_patch_line_stats": {
-          "ignore": true
+          "args": {
+            "total_context": {
+              "isReturn": true
+            },
+            "total_additions": {
+              "isReturn": true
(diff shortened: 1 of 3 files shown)


maxkorp:
Awesome work. I'm assuming just from the unusual data type that the changes to descriptor were required, that you cant just remove the `ignore:true`, correct? Otherwise, nice work!

jfremy:
HI,
if you're asking whether I had to add the parameter in addition to remove the ignore:true part, I think I do. Otherwise, the output from the method won't be returned by the wrapper. At least, that's how I understand it.

johnhaley81:
This looks good to me. Thanks @jfremy!

PR #772, opened 2015-11-05
Add `RevWalk.prototype.fastWalk`

johnhaley81:
This adds the manually written C++ function `fastWalk` the `RevWalk` class.

I also added in a system the let the developer add in optimized C++ code that will be added to
the library in a similar way that we wrap the libgit2 library. Basically, the developer just defines the 
API for the function to wrap and then just writes the C++ implementation of the functions. The
generator classes take care of the rest.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -1807,7 +1807,6 @@
           "isAsync": false
         }
       }
-
     },
 	  "signature": {
 	    "functions": {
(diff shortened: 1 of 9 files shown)


tbranyen:
Ha so awesome, have you noticed a significant performance boost by avoiding the Native->User land bridge?

johnhaley81:
An order of magnitude speed increase :)

johnhaley81:
A few other things on my list are blame and file history. Both of those require a large revwalk and we can do that all in C++ land and just send back up the result. There really shouldn't be too many cases of this but it's a massive perf increase when you need to do it.

PR #776, opened 2015-11-06
Ignore undefined options in normalizeOptions.

kenprice:
It's possible to have some options (DiffOptions, for example) where `options[key]` is undefined. There's no reason why undefined options should be passed to any of nodegit's class methods. Let's check for them in `normalizeOptions`, and prevent these undefined options from being copied.

I didn't check for `null`, since for some options, `null` is ok.

### DIFF
--- a/lib/utils/normalize_options.js
+++ b/lib/utils/normalize_options.js
@@ -18,7 +18,9 @@ function normalizeOptions(options, Ctor) {
   var instance = new Ctor();
 
   Object.keys(options).forEach(function(key) {
-    instance[key] = options[key];
+    if (typeof options[key] !== "undefined") {
+      instance[key] = options[key];
+    }
   });
 
   return instance;
--- a/test/tests/diff.js
+++ b/test/tests/diff.js
@@ -274,6 +274,31 @@ describe("Diff", function() {
       });
   });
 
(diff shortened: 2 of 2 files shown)


tbranyen:
Does this fix a problem? There are few cases where undefined values should matter.

kenprice:
When `options` contains `pathspec = undefined`, and this `options` is passed to `NodeGit.Diff.indexToWorkdir` as its options, this would cause a segfault. 

Of course with proper care in creating `options`, the segfault could be prevented. But on the other hand, properties of `options` shouldn't be undefined, so we should check if it is and ignore any properties that happen to be undefined to prevent side effects like the one I mentioned above. What do you think?

tbranyen:
Makes sense to me, would it be possible to add a unit test with that case illustrated?

kenprice:
Yes I can do that. Will tag as [WIP] in the meantime.

kenprice:
@tbranyen Sorry it took a while. I replicated the exact bug in a new spec test for `Diff`. Behavior tested pretty much is: `indexToWorkDir` should be able to accept undefined options without dying.

tbranyen:
Awesome, no worries man! Thanks for adding it :D

PR #777, opened 2015-11-06
Node v5.0 (modules v47) to allow prebuilt binaries in S3.

samkelleher:
Building nodegit locally in the absence of a prebuilt binary is still a complicated task 😰. By adding Node v5 support to the CI builds now we can get the prebuilt up on S3 quickly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
   matrix:
     - export NODE_VERSION="0.12"
     - export NODE_VERSION="4.1"
+    - export NODE_VERSION="5.0"
 matrix:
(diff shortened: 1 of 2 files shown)


tbranyen:
Makes sense to me, thanks!

samkelleher:
Thanks @tbranyen  - should these same changes also be backported to the 0.5.0 tag - I believe it would build against v5 node, otherwise projects running now with node v5 don't have a prebuilt binary until a new nodegit release is made.

maxkorp:
We can also manually publish those for 0.5.0

PR #780, opened 2015-11-10
Build both x86 and x64 on appveyor

maxkorp:
What it says above. Needed so we can get all the binaries out there.


tbranyen:
Pro move!

saper:
:100:

PR #783, opened 2015-11-11
Allow node-pre-gyp for Electron and node-webkit

joshaber:
As of https://github.com/mapbox/node-pre-gyp/pull/175, node-pre-gyp now understands “electron” as a runtime.

This doesn’t actually create and publish Electron builds—I’ll work on that next—but it does mean that nodegit will try to use a prebuilt binary and then simply fallback to source.

(Context: we’re investigating bundling nodegit in Atom, but CI times explode without nodegit binary support: https://github.com/atom/atom/pull/9213.)

### DIFF
--- a/lifecycleScripts/install.js
+++ b/lifecycleScripts/install.js
@@ -23,10 +23,6 @@ return whichNativeNodish("..")
       console.info("[nodegit] Must build for node-webkit/nw.js");
       return prepareAndBuild();
     }
-    else if (asVersion) {
-      console.info("[nodegit] Must build for atom-shell");
-      return prepareAndBuild();
-    }
     if (fs.existsSync(local("../.didntcomefromthenpmregistry"))) {
       return prepareAndBuild();
     }
--- a/lifecycleScripts/install.js
+++ b/lifecycleScripts/install.js
@@ -19,10 +19,6 @@ return whichNativeNodish("..")
     asVersion = results.asVersion;
   })
   .then(function() {
-    if (nwVersion) {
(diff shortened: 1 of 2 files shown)


maxkorp:
Ha, I've actually been working on exactly this myself. Worth note that at the same time, it doesn't need to custom build for nwjs either, as that's already a valid runtime in node-pre-gyp as well.

johnhaley81:
@joshaber fyi you can run the test suite via `npm run test` and that will check linter errors as well.

johnhaley81:
Also this is amazing.

:heart:

joshaber:
>Worth note that at the same time, it doesn't need to custom build for nwjs either, as that's already a valid runtime in node-pre-gyp as well.

Oh great point :+1: Removed that too.

joshaber:
>Ha, I've actually been working on exactly this myself

@maxkorp Awesome :metal: Have you looked into packaging the binaries for Electron? It's not clear to me yet how all those pieces fit together.

joshaber:
This is ready for review now, I think :pray:

johnhaley81:
So this overall looks good to me. Something we've needed for a while. It would be helpful if we could also pass in arguments to force a build against a specific version of Electron or nwjs but that is really more for dev and custom build environments.

As long as we can publish binaries that are named correctly this should work. :+1:

joshaber:
>It would be helpful if we could also pass in arguments to force a build against a specific version of Electron or nwjs but that is really more for dev and custom build environments.

Absolutely. I have an incoming PR related to that.

PR #784, opened 2015-11-11
[WIP] Build Electron and nwjs binaries

joshaber:
Build nwjs and Electron versions.

This is really hard to test, but as best as I can tell, it Does The Right Thing®.

(This is a continuation of the work in https://github.com/nodegit/nodegit/pull/783.)


joshaber:
Hmm actually the `package` step won't do the right thing still. I'll re-open in a bit.

lifecycleScripts/install.js
```diff
@@ -93,19 +95,24 @@ function build() {
...
+    runtime = "--runtime=electron";
+
     target = "--target=" + asVersion;
 
     distUrl = "--dist-url=https://gh-contractor-zcbenz.s3." +
       "amazonaws.com/atom-shell/dist";
   }
   else if (nwVersion) {
-    builder = "nw-gyp";
+    runtime = "--runtime=node-webkit";
```
joshaber: Given the runtime, `node-pre-gyp` will choose the right builder for us.

maxkorp:
If you want, you can just leave PRs open, and tag them [WIP] in the title. 
I was thinking we may want to try looking for the same vars as node-pre-gyp in the npm config (which will get passed down via the scripts IIRC) rather than trying to determine using which-native-nodish, which was made before any of that stuff got standardized at all.

maxkorp:
So my first thought here is that I don't know if I like reusing the installscript for packaging, because that also will trigger a reconfiguration of libssh2, and an additional regeneration of all of the C++ and gyp files etc. Nothing should have changed, but it seems a bit unneccesary. Factoring it out into some sort of node-pre-gyp helper might be a bit more on point, but that should be a simple transition, and most of your work should just drop right over. 

As an aside, any chance you can interactive-rebase or cherry-pick around that merge and just rebase this sans merge-commit on top of master? 

Awesome friggin work so far man.

joshaber:
>I was thinking we may want to try looking for the same vars as node-pre-gyp in the npm config (which will get passed down via the scripts IIRC) rather than trying to determine using which-native-nodish, which was made before any of that stuff got standardized at all.

Yeah I'd love to get a second set of :eyes: on this! So far as I can tell, `node-pre-gyp` looks at the process's version to see if it's node-webkit or not: https://github.com/mapbox/node-pre-gyp/blob/34224b4264db52441f94d4a91ed0b2d28cadf762/lib/util/versioning.js#L252 (which doesn't make a ton of sense to me?). Otherwise it relies on the `--runtime` flag. So I *think* `which-native-nodish` is still a legit way of checking :+1: 

>So my first thought here is that I don't know if I like reusing the installscript for packaging, because that also will trigger a reconfiguration of libssh2, and an additional regeneration of all of the C++ and gyp files etc.

I agree that's gross :+1: I'll change it.

johnhaley81:
Yeah, it's trying to deduct what platform it's being run in which I don't think we can assume in a lot of build environments. I feel like forcing that in `node-pre-gyp` might be the way to go. We can grab that from `which-native-nodish` or through an argument passed in.

maxkorp:
yeah, just factoring it out though would rock, since the --runtime figuring, and the calling out to it etc are all shared. 

I need to do a little work on which-native-nodish at minimum, but it seems thats fairly non standard the way I implemented it. It currently just recurses up the tree looking for ../.., where the first parent is a node_modules, until it cant go up anymore, then looks in the root most modules package.json to see if its got electron/atom-shell or nwjs/node-webkit in it's engines. I'd like to see it pull from the root-most levels npm configuration (so an npmrc in the root folder, or the users, or the global npm installations config) programatically, since that's basically what node-pre-gyp will do automatically for most modules if my understanding of how they detect that is correct. IIRC the runtime argument works for an override. Not sure if the same holds true at buildtime however. That would super simplify things though.

lifecycleScripts/install.js
maxkorp: You should be able to nuke the builder argument entirely and just hardcode it to node-pre-gyp I think.

joshaber:
Alright, I've pulled apart the functionality into distinct installation and package scripts. Lemme know how you all feel about it.

As for determining runtime, lemme see if I understand you right. `node-pre-gyp` can pull the runtime from npm config, so we should too. So that means either updating `which-native-nodish` or doing it with `nodegit` itself. Does that sound right? Do you think that's necessary for this PR?

maxkorp:
No, I'd argue lets put that in a separate pr, but I'd like to get it out asap. Sorry, just thinking out loud with all of this.

joshaber:
I think this is in pretty good shape, but I'm not sure what's going on with the Appveyor builds. Any theories?

maxkorp:
Huh. I mean, it's only really upset at the electron version ones. Have you tried it locally on a windows machine? I can give it a shot, or rdp into appveyor if needed.

maxkorp:
Weird, so, it dies because node isn't able to require the module, but the ABIs don't missmatch so it doesn't get angry. Checked it out at 87480471258296d73a337ebdb37bac6f71d79885. So when the tests run, they die, because it just "cant find" the module, rather than getting a version mismatch. Gotta make the tests run with electron as well in that case. I know we can make electron behave all node-y rather than opening a window using `ATOM_SHELL_INTERNAL_RUN_AS_NODE="1"`. Unusual that it works in osx/linux though, although that may be related to the executable name or something. I know that was an issue a while back with even renaming atom-shell.exe, so to have some require mismatch causing an error isn't that shocking. .

maxkorp:
Ok cool, so if we just factor out the test script to choose the proper executable (electron vs node) and make sure that env var is set, we _should_ be ok.

maxkorp:
Wtf, it can't find electron?

joshaber:
Seems pretty likely it's a path issue. Maybe we could install it non-globally and call it directly?

maxkorp:
That would work although I'm concerned about why it's not on our path on appveyor. We may just need to add our npm global modules locaton to the path in the appveyor.yml

joshaber:
>We may just need to add our npm global modules locaton to the path in the appveyor.yml

But the weird thing is that we install `node-pre-gyp` the same way, and (AFAIK, at least) it works?

johnhaley81:
So the path in windows is resolving to the root `nodegit` folder. Instead of `nodegit/node_modules/.bin/electron` it gets `node_modules/.bin/electron`.

However fixing that still gives me an error =/

joshaber:
Well, maybe it's time to try using Chocolatey to install it.

johnhaley81:
Got it. Fix incoming.

joshaber:
Oh nice :metal:

johnhaley81:
Ack... don't have write permissions...

joshaber:
Ha, I'll add you.

joshaber:
You should be good now.

johnhaley81:
Ahhh yes... works locally now but not on CI >.<

maxkorp:
:dancer:

maxkorp:
So, a question occured to me last night around 3 during the prolonged affair of "dogs going outside". Does anybody know how we run the tests with nwjs?

johnhaley81:
Right now we don't. But we should probably add that in :)

johnhaley81:
Damn, you ninja'ed my commit.

johnhaley81:
Was about to push that exact change :)

joshaber:
:metal: Thanks @johnhaley81!

So it looks like it's working now, though test output isn't being redirected on Windows. Because reasons, I guess?

maxkorp:
So, piping stdio via spawn is tough, because of exactly that, output going missing on appveyor. We have the same issue in our other project. I think doing a `task.stdout.on('data', process.write)` works, but I'm not even sure of that. Definitely something we need to fix though.

maxkorp:
Wait, dc27e1015b7ff61517906823a56130dc4e974f65 is showing proper test output after all. Are you not seeing it?

joshaber:
I don't see it in https://ci.appveyor.com/project/TimBranyen/nodegit/build/2013/job/llb8xttbf301j39s?

maxkorp:
O.O it's only broken in electron.

joshaber:
>O.O it's only broken in electron.

Ah nice! That's less bad. I'll fool around a bit more, but that certainly makes it sound like Electron's fault and not the changes.

maxkorp:
So as a side note, I figured out the global install issue. If you change the command on windows to "electron.cmd" it works. On the cli, `electron` works, but its falling back to the second value of `where electron` which yields `electron.cmd`. Using that in the spawn works. Odd, but it makes sense at least. I know npm scripts used to suffer from the same issue.

maxkorp:
Yeah our other project with that problem is an electron project, so it makes sense.

maxkorp:
Can I grab push access really quickly? want to try a few things with appveyor.

joshaber:
>Can I grab push access really quickly? want to try a few things with appveyor.

:boom: Thought at this point the bottleneck is how quickly appveyor can build :smile:

maxkorp:
Yeah, I'm going to start cancelling builds as soon as we know if it worked or not lol.

joshaber:
Awesome, thanks for the help!

maxkorp:
No problem. This was literally the top thing on my plate anyways once I got another project done yesterday, so having 99% of it already done is nothing I'm going to complain about. 

I suppose I can start poking at the tests with nwjs while we wait :/

appveyor.yml
```diff
@@ -44,16 +46,19 @@ install:
   - npm install
 
 test_script:
+  - cmd: set PATH=%APPDATA%\npm;%PATH%
+  - cmd: set PATH=%APPDATA%\Roaming\npm;%PATH%
+  - cmd: npm install -g electron-prebuilt@%ELECTRON_VERSION%
```
kevinsawicki: I thought `cmd:` was the default so you might not need the prefixes here.

kevinsawicki:
This is going to be awesome :fireworks:

appveyor.yml
joshaber: Yeah I think you're right. There's definitely some cleanup to do once this is working :+1:

maxkorp:
Yeah I think if it's prefixed he actually calls like `cmd \s "cdm {line contents}"` or something, so just trying to trim stuff down. I think a few of the commands did require CMD though, so they were added throughout to keep reading consistent. May not be required at all anymore, but that's a separate issue. We just gotta figure out this logging problem and then do some squashing and this is good to go.

maxkorp:
Hrm, no, something else seems wrong. It apparently takes <1 sec to run the tests in electron, but >4 minutes in node? That seems... incorrect.

joshaber:
Yeah, though curiously in https://ci.appveyor.com/project/TimBranyen/nodegit/build/2015/job/q2m3y0udamiqvj3h they seem to have run (elapsed time ~3 minutes in testing).

lifecycleScripts/runTests.js
```diff
@@ -0,0 +1,41 @@
...
+var path = require("path");
+
+var determineTarget = require("./determineTarget");
+
+return determineTarget()
+  .then(function(targetInfo) {
+    var opts = {
+      cwd: ".",
+      maxBuffer: Number.MAX_VALUE,
+      env: process.env,
```
kevinsawicki: You might want to copy `process.env` here or else when you set `opts.env.ATOM_SHELL_INTERNAL_RUN_AS_NODE` down below it will actually be setting `process.env.ATOM_SHELL_INTERNAL_RUN_AS_NODE` as well I believe.

maxkorp: I was going to ask, does the `cwd: "."` get us anything here?

maxkorp:
@FeodorFitsner Any ideas why logging is not coming through here in the case of electron? I know the output gets redirected through chromiums logger. Is it plausible this is getting stripped somehow by one of your processes?

FeodorFitsner:
I don't have an clue. AppVeyor does nothing special with process output.

maxkorp:
No sanitization or anything? Shoot. 

I've got a super simple example that works locally, but in theory should break on appveyor, so I'll hook it up in a new repo and see if it breaks.

maxkorp:
Also, not going to worry about automating nw.js with all of this for now, that's a huuuuuge can of worms apparently.

joshaber:
Alright, we're now failing properly on Travis and Appveyor when the tests failed :point_up:

johnhaley81:
:+1:

maxkorp:
Nice :+1:

johnhaley81:
So I think this is almost ready. I'd say lets clean up the commit history since it's like 30 commits of just shooting in the dark with what's going on with CI. Squash those suckers into something more succinct and then let's merge this!

johnhaley81:
So it's dying on linux consistently but tbh I'm not sure why. It looks like a seg fault since the tests just bomb out but before they bomb out there are some errors.

joshaber:
>If we're going to be bringing in methods like these, might be best to standardize on something like lodash's modular packages.

:+1: 

At this point it seems like we've solved a lot of the systematic problems, but we're hitting a lot of test races. So I'll poke around and see if there's anything eminently fixable.

johnhaley81:
So @maxkorp and I talked about this a bit and we think it's because the other native modules that are used are build against node headers instead of electron. That could be related to this really weird situation.

Also we're using this in GitKraken and I can confirm that the tests that are failing are indeed working in GK.

maxkorp:
Denied >.<

maxkorp:
I'm starting to seriously wonder if `ATOM_SHELL_RUN_AS_NODE` isn't really short for `ATOM_SHELL_RUN_IN_TERMINAL_BAD`. 

I'm going to try importing these tests into our app that we know uses nodegit successfully, so they can run in electron in a normal way.

joshaber:
Superseded by #804. :metal:

PR #789, opened 2015-11-18
Bump nodegit-promise to 4.0.0 and Promisify to 0.3.0

Mr-Wallet:
Fixes issues where promises that resolve to promise chains could end up in an invalid state where they were neither pending, fulfilled, nor rejected.


maxkorp:
Whoops, I have this in a commit already, forgot to PR it. Just nodegit-promise isnt enough, have to bump promisify too.

johnhaley81:
#790 and this collide.

PR #792, opened 2015-11-20
TravisCI manifest cleanup

Turbo87:
This PR cleans up the TravisCI manifest file. e.g. installing `nvm` on Linux is unnecessary because it's preinstalled, lcov should only be installed if it is actually needed, ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,26 @@
+branches:
+  only:
+    - master
+    - /^v\d+\.\d+\.\d+$/
+
+os:
+  - linux
+  - osx
+
 sudo: false
+
+env:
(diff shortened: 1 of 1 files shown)


tbranyen:
Awesome! Thank you :+1:

Turbo87:
@tbranyen maybe I'm missing something, but where did those changes go? I can't find them in https://github.com/nodegit/nodegit/blob/master/.travis.yml anymore?!

johnhaley81:
Argh, that was my fault. I'll fix it up...

johnhaley81:
Everything should be ok now. Sorry about that =/

PR #796, opened 2015-11-23
bump openssl to match nodejs' version

implausible:
By bumping openssl to what nodejs is building with, we are able to patch a problem with electron compilation that prevents nodegit from connecting to anything over https in linux on older systems (12.04).


tbranyen:
Pretty sure this won't work in Windows.

implausible:
Why won't it work in windows?

tbranyen:
OpenSSL is a requirement for libssh which is a requirement for libgit2 on any operating system. We definitely use and build it in Windows. We include it as a vendor dependency for that very reason.

tbranyen:
Although I think I'm misunderstanding the purpose of this PR. Are you simply bumping it to the version Node uses? I can't get the diff to load right now without crashing my tab.

implausible:
Yes, updating to the latest version of openssl so that we can use the gyp file for it. There's an issue with compilation with nodegit and electron that is solved with this gyp file / copy of openssl.

PR #802, opened 2015-11-26
Added missing arguments to compare

### LINKED ISSUE #800: Bug: False positives when using compare
The following methods give false positives when comparing unequal objects:
- `Hashsig#compare`
- `Oid#cmp`
- `Oid#equal`
- `Oid#ncmp`
- `Reference#cmp`

smith-kyle:
This should fix #800. The issue was that the second argument was marked as `isSelf`in the following methods. So comparing two objects really just compared an object to itself.

From what I could tell this will only change the following:
- `Hashsig#compare`
- `Oid#cmp`
- `Oid#equal`
- `Oid#ncmp`
- `Reference#cmp`
- `Strarray#copy`


maxkorp:
Awesome, thanks!

PR #804, opened 2015-12-02
[WIP] Use node-pre-gyp-npmrc-helper for building

maxkorp:
This gets rid of all the crazy shit to build for electron/nwjs, and just references an npmrc file (or any npm config really) when a parent project/module installs nodegit. 

This is going to require publishing some beta versions to test unfortunately >.<


joshaber:
Fantastic :sparkling_heart:

PR #805, opened 2015-12-02
Force read index when performing rebase

srajko:
The index is read to make sure the conflict check is performed on the true current state of the index.

Without this, I've seen problems where `hasConflicts` would report conflicts that were not really there.

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -880,6 +880,7 @@ function performRebase(repository, rebase, signature) {
     .then(function(rebaseOperation, lel) {
       return repository.openIndex()
         .then(function(index) {
+          index.read(1);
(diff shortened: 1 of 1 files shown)


maxkorp:
Any way to add tests for this? Or is the issue unreliable in that regard?

srajko:
I can try.

johnhaley81:
http://imgur.com/gallery/WOpf0Sf

srajko:
Good news - I was able to write a test that breaks the rebase.  Bad news - this fix doesn't make the test pass.  It takes a few more index force-reads to make it pass, at which point I'm not so sure this is the right direction.

I will close this PR and submit something different.

PR #807, opened 2015-12-05
Optional options for Blame.file

### LINKED ISSUE #808: Options should be optional in Blame.file
According to the libgit2 docs if the `options` aren't passed in default to a newly initialized `BlameOptions`

smith-kyle:
This should fix #808

### DIFF
--- a/lib/blame.js
+++ b/lib/blame.js
@@ -1,0 +1,18 @@
+var NodeGit = require("../");
+var Blame = NodeGit.Blame;
+var _file = Blame.file;
+var BlameOptions = NodeGit.BlameOptions;
+
+/**
+ * Retrieve the blame of a file
+ *
+ * @param {Repository} repo that contains the file
+ * @param {String} path to the file to get the blame of
+ * @param {BlameOptions} [options] Options for the blame
+ */
+Blame.file = function(repo, path, options) {
+  if (!options) {
+    options = new BlameOptions();
+  }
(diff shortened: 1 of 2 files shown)


lib/blame.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Retrieve the blame of a file
+ *
+ * @param {Repository} repo that contains the file
+ * @param {String} path to the file to get the blame of
+ * @param {BlameOptions} [options] Options for the blame
+ */
+Blame.file = function(repo, path, options) {
+  if (!options) {
```
johnhaley81: Do https://github.com/nodegit/nodegit/blob/master/lib/checkout.js#L18 instead please.

smith-kyle:
It looks like `normalizeOptions` will return null if `!options` is `true` but I'm trying to instantiate options if they aren't passed in. Maybe something like:
```javascript
options = normalizeOptions(options)
if (options === null) {
  options = new BlameOptions();
}
```

johnhaley81:
libgit2 should handle the null case. If it's null libgit2 creates the default. We shouldn't need to pass it in. At least we don't in other areas. Is this different?

smith-kyle:
Thanks for that. I don't know why I didn't think of libgit2 taking care of the default :+1:

johnhaley81:
np :+1:

PR #812, opened 2015-12-07
Let node-pre-gyp handle Electron/nwjs

joshaber:
`node_pre_gyp` [will parse `npm_config_` environment variables](https://github.com/mapbox/node-pre-gyp/blob/cfefa6f6197748d50980c5e49dc57c44ef3efc70/lib/node-pre-gyp.js#L130) to determine if the runtime is node/electron/nwjs. So nodegit doesn't need to do any special detection itself.

I believe this supersedes #804.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -15,7 +15,6 @@
 .jshintrc
 .travis.yml
 appveyor.yml
-.didntcomefromthenpmregistry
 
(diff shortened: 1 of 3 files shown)


joshaber:
Replaced by #815.

PR #819, opened 2015-12-09
TreeBuilder.create has source as optional param, plus associated tests

petemill:
TreeBuilder.create(repo, null) should be allowed, since this is the method required to create new directories in any tree, or the initial tree for the initial commit of a repo.

In this branch I've specified that `source` is optional (though you still need to send `null` or it will throw `Error: Callback is required and must be a Function` - @maxkorp stated that this is a semi-known issue with the promisifier).

The tests are an initial set for Treebuilder: check it can create a new Tree, and check it can be created from an existing tree and verify it has the correct number of entries. I plan to add more tests verifying it can add and read specific types of entries, but that's unrelated to this fix.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2099,6 +2099,13 @@
               "isReturn": true
             }
           }
+        },
+        "git_treebuilder_new": {
+          "args": {
+            "source": {
+              "isOptional": true
+            }
+          }
         }
       }
     },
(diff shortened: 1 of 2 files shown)


johnhaley81:
Looks good. Thanks @petemill!

maxkorp:
I was thinking that last night. Mucho idea-o likeo

PR #820, opened 2015-12-09
enable access to git_index_new and add test

petemill:
Test is intermittently failing right now. The new index, which does not have a repository, and only 1 entry added to it, has an entry in a conflict state, but only sometimes when it is run.

When doing `index.conflictGet` only 1 of `ours`, `theirs`, `base` has the entry we added, the other 2 are null.

Not sure why - would appreciate any ideas.


johnhaley81:
Can you rebase this on top of master and enable thread safety to see if that fixes the issue please?

petemill:
@johnhaley81 rebased but unfortunately test still fails intermittently (fails more often than succeeds).

I ran the tests both with `NODEGIT_TEST_THREADSAFETY=true` and with `NodeGit.enableThreadSafety()` explicitly inside the test. Fails on both occasions with `Index should not have conflicts`.

johnhaley81:
@petemill Alright! Let's try this again. Can you rebase on top of #971 and see if that helps?

PR #821, opened 2015-12-09
Add submodule foreach

joshaber:
A lot of this is to work around the fact that `git_submodule_foreach`’s callback isn’t `typedef`’d. That’s fixed in https://github.com/libgit2/libgit2/pull/3539, but there hasn’t been a release including it yet. So in the mean time, we need to do a little extra futzing.


johnhaley81:
LGTM

PR #825, opened 2015-12-10
More natural API for getting submodule names.

joshaber:


lib/repository.js
```diff
@@ -1094,6 +1095,23 @@ Repository.prototype.getStatusExt = function(opts) {
...
+ * @return {Array<String>}
+ */
+Repository.prototype.getSubmoduleNames = function() {
+  var names = [];
+  var callback = function(submodule, name, payload) {
+    names.push(name);
+  };
+
+  return Submodule.foreach(this, callback).then(function() {
+    return names;
```
johnhaley81: We also support using callbacks as well. Reference https://github.com/nodegit/nodegit/blob/master/lib/repository.js#L98-L100 for an example on implementing.

johnhaley81:
@joshaber just 1 quick comment and then we're GTG IMO.

johnhaley81:
LGTM. Thanks @joshaber

Issue #827, opened 2015-12-11
Threading model


joshaber:
:wave: 

I've been plumbing the depths of nodegit to understand how all the pieces fit together. I'm a bit concerned about the threading model.

My understanding:

1. We enqueue asynchronous work with [`Nan::AsyncQueueWorker`](https://github.com/nodegit/nodegit/blob/c4292de4adb091c1fa7daed8515c105776714310/generate/templates/partials/async_function.cc#L77).
2. `Nan::AsyncQueueWorker` calls [`uv_queue_work`](https://github.com/nodejs/nan/blob/367e82a60fbaa52716232cc89db1cc3f685d77d9/nan.h#L1702-L1709).
3. `uv_queue_work` [enqueues the work on a threadpool](http://docs.libuv.org/en/v1.x/threadpool.html#c.uv_queue_work).

This means:

1. We could do work on multiple threads at once.
2. libgit2 says its objects [shouldn't be used across multiple threads at once](https://github.com/libgit2/libgit2/blob/6aa06b65cfcce164d7ad3ef184f40acb68406078/THREADING.md#sharing-objects).
3. We could end up accessing the same objects across multiple threads at once, most often the repository object.
4. Potential :boom:, definite violation of libgit2's threading model.

Is my understanding correct? Am I missing something?

srajko:
Heya,

I've also been looking into this.  You are right about `Nan::AsyncQueueWorker` and that calling `uv_queue_work`, but afaict the call queues the call on the `uv_default_loop()` (https://github.com/nodejs/nan/blob/master/nan.h#L1704), which is supposed to be the default node.js loop (https://nikhilm.github.io/uvbook/basics.html#default-loop).  So I don't think it's using a threadpool, but I might be wrong.

However, even if it is using an event loop on a single thread, this is only for NodeGit's async functions - the sync ones are presumably executing on a different thread, which would lead to the same potential violation of libgit2's threading model.

joshaber:
@srajko I think that's the loop that the `after_work_cb` is delivered on, so that you end up back on the default event loop when you're done.

But the `work_cb` is called on the thread pool. I _think_. Which is what introduces the asynchrony. I _think_. The docs aren't terribly clear :disappointed:

joshaber:
Prior art: https://github.com/mapbox/node-sqlite3/blob/master/src/database.cc is more or less solving this same problem.

johnhaley81:
[...]

PR #829, opened 2015-12-11
[WIP] Call libgit2 from a dedicated thread

srajko:


saper:
Ouch, this is going to be tough one:)

johnhaley81:
I don't think we need this anymore after #827.

PR #830, opened 2015-12-12
Bump libssh2 to 1.6.0

maxkorp:


tbranyen:
Looks like this is outdated now, since master is on 1.6.0

silenio:
It seems master is still on libssh2 1.4.3. Should this issue be reopened?

maxkorp:
Yeah, looking at the libssh2 directory, the current libssh2.h file states it's version is 1.4.3

PR #832, opened 2015-12-14
Set the npm_config_ runtime and target too.

joshaber:
I *believe* this is at least One Of The Problems with the Windows builds. We’re a little too eager to download a binary instead of building.

1. Our [install script](https://github.com/nodegit/nodegit/blob/1a07da394b646e6c382c78998e65a0b4cfcf0f54/lifecycleScripts/install.js#L13) sees `include` and `src` since we downloaded a tarball of the repository.
2. We mark `fromRegistry` as `true`.
3. So we try to install a prebuilt.
4. Good news everyone! There’s already a binary up for `nodegit-v0.6.0-node-v14`!
5. We download the binary and think our job is done, instead of building.

It’s debatable whether this is the bug, or the bug is that we think we’re from the registry when we downloaded a tarball. :shrug:

### DIFF
--- a/lifecycleScripts/install.js
+++ b/lifecycleScripts/install.js
@@ -18,6 +18,24 @@ catch(e) {
   fromRegistry = false;
 }
 
+var electronVersion = process.env.ELECTRON_VERSION;
+var nwjsVersion = process.env.NWJS_VERSION;
+
+process.argv.forEach(function(arg) {
+  if (~arg.indexOf("electronVersion")) {
+    process.env.npm_config_runtime = "electron";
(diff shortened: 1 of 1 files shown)


joshaber:
Gah, false alarm. `include` and `src` *shouldn't* be there.

maxkorp:
They _should_ when installing a published build (since they're allowed via npmrc), but not on the CI servers.

johnhaley81:
lol https://github.com/nodegit/nodegit/pull/832#issuecomment-164341629

PR #838, opened 2015-12-18
Omit baton->error on GIT_ITEROVER

srajko:
This addresses a bug that would happen occasionally during rebase - because `Rebase.next` doesn't set or clear `giterr_last` when it returns `GIT_ITEROVER`, we would pick up a stale error if one was previously left in that thread.  `performRebase` would then get the stale error instead of `GIT_ITEROVER` and not finish correctly (https://github.com/nodegit/nodegit/blob/d1201998a6fe64e35cf8c357b3dd94d2f6982bc0/lib/repository.js#L904-L908).

I am not sure whether this is the best way to fix this, and if it is, whether similar adjustments need to be made in other places in code.  For example, https://github.com/nodegit/nodegit/blob/d1201998a6fe64e35cf8c357b3dd94d2f6982bc0/generate/templates/partials/sync_function.cc#L66-L70 accesses `giterr_last` where `result` could be `GIT_ITEROVER`.

An alternative to this fix might be to do a `giterr_clear` before we call into `libgit2`, for example these places: https://github.com/nodegit/nodegit/blob/d1201998a6fe64e35cf8c357b3dd94d2f6982bc0/generate/templates/partials/sync_function.cc#L38 https://github.com/nodegit/nodegit/blob/d1201998a6fe64e35cf8c357b3dd94d2f6982bc0/generate/templates/partials/async_function.cc#L82-L86.  That way `giterr_last` won't be stale to begin with, and we might not even need to have special handling for `GIT_ITEROVER`.

### DIFF
--- a/generate/templates/manual/revwalk/fast_walk.cc
+++ b/generate/templates/manual/revwalk/fast_walk.cc
@@ -30,6 +30,7 @@ void GitRevwalk::FastWalkWorker::Execute()
   for (int i = 0; i < baton->max_count; i++)
   {
     git_oid *nextCommit = (git_oid *)malloc(sizeof(git_oid));
+    giterr_clear();
     baton->error_code = git_revwalk_next(nextCommit, baton->walk);
 
     if (baton->error_code != GIT_OK)
--- a/generate/templates/partials/async_function.cc
+++ b/generate/templates/partials/async_function.cc
@@ -79,6 +79,7 @@ NAN_METHOD({{ cppClassName }}::{{ cppFunctionName }}) {
 }
 
 void {{ cppClassName }}::{{ cppFunctionName }}Worker::Execute() {
+  giterr_clear();
   {%if .|hasReturnType %}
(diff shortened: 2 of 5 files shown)


johnhaley81:
@srajko I feel like clearing the error is the better approach to this. At least for how we use libgit2 in NodeGit. Can you try that out in both sync and async and see if that fixes it?

srajko:
@johnhaley81 I like that one better too.  Tried it out and it also fixes the bug :-).

The only weird thing was that calling `giterr_clear` would crash in `global.c`, so I added a call to `git_libgit2_init` which we were supposed to be calling anyway (https://libgit2.github.com/libgit2/#HEAD/group/libgit2/git_libgit2_init).

srajko:
Ah, I found where we were calling `init`.  I will take that one out since it needs to be called sooner (I think what was happening is `exports.Libgit2.init()` would call `giterr_clear` before `git_libgit2_init`)

johnhaley81:
:zap: :+1: :zap:

PR #841, opened 2015-12-22
Use libgit2 fork to fix ssh url detection

johnhaley81:
This should add support for the ssh+git:// and git+ssh:// protocol schemes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   },
   "vendorDependencies": {
     "libgit2": {
-      "sha": "e8feafe32007ebd16a61820c70abd221655d053c",
+      "sha": "c2786d822cfafe40c0b27392d0b29ece0c426e36",
       "version": "0.23.4"
     },
     "libssh2": "1.6.0",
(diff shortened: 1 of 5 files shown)


johnhaley81:
I don't want this merged until we have https://github.com/libgit2/libgit2/pull/3555 merged. I definitely don't want to deviate from libgit2 and only want to pull this in as an advanced copy of the fix.

maxkorp:
> deviate from libgit2
:shudder:

tbranyen:
Is this PR still necessary @johnhaley81 ?

johnhaley81:
@tbranyen so there was a bit of debate on whether or not this should be included since it's actually wrong URL formating. The issue is that somehow it got into git-core and people migrating from svn must have thought that svn+ssh carried over to git and started formatting their URL's like that. However it happened it's now a supported protocol format according to git-core so we (both libgit2 and NodeGit) have to support it as well.

I was waiting for the parent PR to get merged into libgit2 before I merged this as well since I'm trying to not maintain a deviated fork of libgit2. That just happened this morning so libgit2 now supports it although not in the latest (0.24.0) release. As far as I can tell we're good to merge this now.

johnhaley81:
So https://github.com/libgit2/libgit2/pull/3555 is merged now but I think a better course of action is to just update libgit2. I'm going to close this.

Issue #785, opened 2015-11-12
Length of String is wrong.


watiko:
NodeGit uses substring with bytes length comes from strlen but JavaScript requires characters length.
So non-ascii string will be broken.

https://libgit2.github.com/libgit2/#HEAD/type/git_diff_line
git_diff_line->content_len : Number of bytes of data

NodeGit:
https://github.com/nodegit/nodegit/blob/v0.5.0/lib/convenient_line.js#L71

Some other implementations:
https://github.com/libgit2/rugged/blob/v0.23.3/ext/rugged/rugged_diff_line.c#L68
https://github.com/libgit2/libgit2sharp/blob/v0.21.1/LibGit2Sharp/Core/EncodingMarshaler.cs#L129
https://github.com/libgit2/objective-git/blob/0.8.8/ObjectiveGit/GTDiffLine.m#L22

Thx.

maxkorp:
Good catch!

watiko:
I wrote a following ugly fix in the layer of JavaScript.
But I think this issue should be resolved in that of C++.

```diff
--- a/convenient_line.js	2015-11-13 23:47:56.299338415 +0900
+++ b/convenient_line.js	2015-11-13 23:48:19.766911535 +0900
@@ -67,8 +67,22 @@
 * @return {String}
 */
 ConvenientLine.prototype.content = function() {
-  return this.raw.content()
-    .substring(0, this.raw.contentLen()).replace("\n", "");
+  function countBytes(chr) {
+      return (encodeURIComponent(chr).replace(/%../g,"x").length);
+  }
+
+  var rawContent = this.raw.content();
+  var rawContentLen = this.raw.contentLen();
+  var trueContentLen = 0;
+
+  rawContent.split('').some(function(chr) {
+    rawContentLen -= countBytes(chr);
+    trueContentLen += 1;
+    if (rawContentLen == 0)
+      return true;
+  });
+
+  return rawContent.substring(0, trueContentLen).replace("\n", "");
 };
 
 NodeGit.ConvenientLine = ConvenientLine;
```

johnhaley81:
@watiko would you mind PR'ing this? The problem with implementing this in C++ is that we'd have to figure out way to generate that code in our generators. I'd argue that you *should* be using the JS convenience wrappers/functions unless you really know what you're doing and want that freedom.

Also, could you cache this value? We should need to recalculate this stuff since it's immutable.

watiko:
[...]

PR #844, opened 2015-12-28
Modify `ConvenientLine` to work with unicode

johnhaley81:
This changes how `ConvenientLines` are calculated with a breaking API change.

`ConvenientLine` used to assume that all content coming back was ASCII characters and
a uniform byte length. This is no longer the case. Now we use the byte length coming
back from libgit2 to extract the string from a buffer using the correct byte length. This
will give back the correct line everytime.

~~The breaking API change is that we no longer return the new line character with the line
nor do we include that character in the `contentLen` calculation either. I'm on the fence
about that change but it seems like it's what we ought to be doing instead of force all of
the consumers of the library to account for that everywhere.~~

Edit: I reverted the change to returning the new line characters.

### DIFF
--- a/lib/convenient_line.js
+++ b/lib/convenient_line.js
@@ -3,59 +3,63 @@ var NodeGit = require("../");
 function ConvenientLine(raw, i) {
   this.raw = raw;
   this.i = i;
+  this._cache = {};
 }
 
 /**
-* The content of the line
-* @return {String}
+* The type of the line (context, addition, deletion, etc...)
+* @return {Diff.LINE}
 */
 ConvenientLine.prototype.origin = function() {
  return this.raw.origin();
 };
 
 /**
-* The content of the line
-* @return {String}
+* Line number in old file or -1 for added line
+* @return {Number}
 */
 ConvenientLine.prototype.oldLineno = function() {
  return this.raw.oldLineno();
 };
 
 /**
-* The content of the line
-* @return {String}
+* Line number in new file or -1 for deleted line
+* @return {Number}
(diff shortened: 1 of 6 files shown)


johnhaley81:
This fixes #785

johnhaley81:
cc @nodegit/owners anybody want to chime in on the line ending stuff?

tbranyen:
Seems like a pretty large breaking change, major version bump? If this brings us closer in alignment to libgit2 (I think you're saying it is since we're using the byte length returned), then I'm :+1:.

johnhaley81:
@tbranyen so the byte count calc is definitely in line with what we should be doing. The removal of the `\n` isn't in line with libgit2 however. After thinking about it more I feel like it should probably be left in and let the consumer of the lib deal with it how they want to.

We might want to do a major version bump at this point anyways since it would be nice to adhere to semvar and not break people's code with a release. Especially if we going to start releasing more often.

So I just argued myself out of cutting off the line endings. I'll change that and push it. If there are no objections from @nodegit/owners then I'll go ahead and merge this.

maxkorp:
Wait so did this not already trim the newline here? https://github.com/nodegit/nodegit/commit/74c19f86eb3903b8971a872da54695a7dc60d02e#diff-3af6ec85ce097f4106d101a2f0f41c9cL71

maxkorp:
Ah nvm, it's then manually reappended each time. Now we just get the utf8 string and don't trim that off, but don't re-add it. I'm all for it :+1:

PR #847, opened 2016-01-03
Fixed example walk-history-for-file.js

pevers:
The walk-history-for-file.js example was broken so I fixed it with some minor changes. The reason it broke is that ```diff.patches()``` returns a promise and not an array that can be reduced. I made the example a little bit less efficient (checking for an already present commit) but I think it is a little bit more readable. The screenshot shows a comparison of ```git log``` and the example. Note that it is not a 1-1 mapping because git log is a little bit more intelligent.

```output of git log descriptor.js```
selection_005
```output of walk-history-for-file.js```
selection_006


johnhaley81:
Thanks for the PR @pevers!

PR #848, opened 2016-01-04
Fix pull and clone example fetchOptions structure

nicolasartman:
[This line] (https://github.com/nodegit/nodegit/blob/232dbb19116c58d22cd7afd7435ee0bbe063f85b/lib/repository.js#L688) and my own trial and error with a project leads me to believe the auth callbacks need to be nested under the `callbacks` key in fetch options.

This also [applies](https://github.com/nodegit/nodegit/blob/44779afd0d460011e13a5cba9a47a48119aa472e/lib/clone.js#L31) to the clone example I think, though it requires a structure of fetchOpts.callbacks.<authCallback>.

### DIFF
--- a/examples/clone.js
+++ b/examples/clone.js
@@ -10,11 +10,13 @@ fse.remove(path).then(function() {
     "https://github.com/nodegit/nodegit.git",
     path,
     {
-      remoteCallbacks: {
-        certificateCheck: function() {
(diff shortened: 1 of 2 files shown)


johnhaley81:
Thanks for the PR @nicolasartman!

PR #854, opened 2016-01-06
Eliminate promise polling

srajko:
Eliminates the polling of promises when executing javascript callbacks from within async libgit2 calls. Replaces with a system that forwards the promise result / rejection reason to the native layer directly.

This should remove our dependency on the nodegit-promise library.

### DIFF
--- a/generate/templates/manual/src/promise_completion.cc
+++ b/generate/templates/manual/src/promise_completion.cc
@@ -1,0 +1,103 @@
+#include "../include/promise_completion.h"
+
+Nan::Persistent<v8::Function> PromiseCompletion::newFn;
+Nan::Persistent<v8::Function> PromiseCompletion::promiseFulfilled;
+Nan::Persistent<v8::Function> PromiseCompletion::promiseRejected;
+
+// initializes the persistent handles for NAN_METHODs
+void PromiseCompletion::InitializeComponent() {
+  v8::Local<v8::FunctionTemplate> newTemplate = Nan::New<v8::FunctionTemplate>(New);
+  newTemplate->InstanceTemplate()->SetInternalFieldCount(1);
+  newFn.Reset(newTemplate->GetFunction());
+
+  promiseFulfilled.Reset(Nan::New<v8::FunctionTemplate>(PromiseFulfilled)->GetFunction());
+  promiseRejected.Reset(Nan::New<v8::FunctionTemplate>(PromiseRejected)->GetFunction());
+}
+
+bool PromiseCompletion::ForwardIfPromise(v8::Local<v8::Value> result, AsyncBaton *baton, Callback callback)
+{
+  Nan::HandleScope scope;
+
+  // check if the result is a promise
+  if (result->IsObject()) {
+    Nan::MaybeLocal<v8::Value> maybeThenProp = Nan::Get(result->ToObject(), Nan::New("then").ToLocalChecked());
+    if (!maybeThenProp.IsEmpty()) {
+      v8::Local<v8::Value> thenProp = maybeThenProp.ToLocalChecked();
+      if(thenProp->IsFunction()) {
+        // we can be reasonably certain that the result is a promise
+
+        // create a new v8 instance of PromiseCompletion
+        v8::Local<v8::Object> object = Nan::NewInstance(Nan::New(newFn)).ToLocalChecked();
+
+        // set up the native PromiseCompletion object
+        PromiseCompletion *promiseCompletion = ObjectWrap::Unwrap<PromiseCompletion>(object);
+        promiseCompletion->Setup(thenProp.As<v8::Function>(), result, baton, callback);
+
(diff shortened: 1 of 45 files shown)


johnhaley81:
This looks fantastic! A+ man. Can you switch the clones tests over to using another library to confirm that we work with other promise implementations? Bluebird or then would be good.

tbranyen:
:dancers: !

maxkorp:
I would also test vanilla node promises (since they exist in 0.12+)

srajko:
Good call!  I tested with native promises since that was the easiest, and got it to work.  On the native side I just needed to replace a `Nan::Get` with a `Nan::GetRealNamedProperty`, and on the test side I had to modify some uses of `Promise.resolve` and `Promise.reject`.

It turns out (at least in node 4.2.3 that I was testing with) that native Promises don't let you call `Promise.resolve` or `Promise.reject` without the `Promise` context.  That is, if you try:
```
var resolve = Promise.resolve;
resolve(1);
```
you get a crash.  And we had test code like this:
```
    return Remote.delete(repository, "origin3")
      .then(function() {
        return Remote.lookup(repository, "origin3");
      })
      .then(Promise.reject, Promise.resolve);
```
The last line results in the same crash.  Use cases like this just need to be modified to pass `Promise.reject.bind(Promise)` and `Promise.resolve.bind(Promise)` instead.  That or we `Promise.resolve = Promise.resolve.bind(Promise)`.

Would you like me to replace `var Promise = require('nodegit-promise')` with `var Promise = global.Promise || require('nodegit-promise')` as part of this PR (and fixup the uses of `Promise.resolve` and `Promise.reject` described above)?

srajko:
Just FYI, the last thing I am working on is ensuring that `PromiseCompletion` objects get freed correctly.  I don't think they are currently - even if I force garbage collection, I never seem to get a hit in the destructor.

srajko:
Actually, I may have been wrong about needing `GetRealNamedProperty` instead of `Get`...

srajko:
Yeah... best I can tell, `GetRealNamedProperty` is like `Get` except it doesn't run interceptors.  Going back to `Get`.

srajko:
Found and fixed the memory leak.

johnhaley81:
@srajko `require('nodegit-promise') is scattered throughout the application in both the `/lib` and the `/test` folder. We should remove all of those and then make sure that the tests still run. I think that would be sufficient for this PR to be considered finished.

johnhaley81:
AppVeyor timed out. I'll re-run

generate/templates/manual/src/promise_completion.cc
```diff
@@ -0,0 +1,103 @@
...
+void PromiseCompletion::Setup(v8::Local<v8::Function> thenFn, v8::Local<v8::Value> result, AsyncBaton *baton, Callback callback) {
+  this->callback = callback;
+  this->baton = baton;
+
+  v8::Local<v8::Object> promise = result->ToObject();
+
+  v8::Local<v8::Object> thisHandle = handle();
+
+  v8::Local<v8::Value> argv[2] = {
+    Bind(promiseFulfilled, thisHandle),
```
johnhaley81: I feel like these should bind to `result` and not `thisHandle`. They are instance properties of `result` and we're rebinding the context of them.

johnhaley81:
https://github.com/nodegit/nodegit/pull/854/files#r49623812 _might_ fix the binding issue?

generate/templates/manual/src/promise_completion.cc
srajko: `promiseFulfilled` and `promiseRejected` are instance properties of `PromiseCompletion` objects - https://github.com/srajko/nodegit/blob/8dae154aa00b5c23f482b746ba4efcfb524a812d/generate/templates/manual/src/promise_completion.cc#L13-L14 https://github.com/srajko/nodegit/blob/8dae154aa00b5c23f482b746ba4efcfb524a812d/generate/templates/manual/src/promise_completion.cc#L97-L103

We need to bind `thisHandle` to them so that they can access the `PromiseCompletion` instance here - https://github.com/srajko/nodegit/blob/8dae154aa00b5c23f482b746ba4efcfb524a812d/generate/templates/manual/src/promise_completion.cc#L92-L94

There might be a better way to do the binding though - using the JS `Function.bind` method was the best I could come up with, short of leaking memory by creating a function template for each instance of `PromiseCompletion` :-)

johnhaley81: So I was mistaken about what was going on. I thought this was related to how the `Promise.resolve` functions passed in now were being bound to themselves in JS. Apparently that is a requirement of Chromium and is Promise library implementation specific. The actual `then` library does not have this requirement and that was the backing lib for `nodegit-promise` which is why we didn't need to bind the functions being passed in during tests. With that resolved I'm good with this PR.

PR #855, opened 2016-01-06
Update to OpenSSL 1.0.2e

johnhaley81:
This is needed because the AVX processor compatibility check can kill some versions of linux that are run on a processor that doesn't have that feature and will throw a SIGILL.

Related: https://github.com/nodejs/node/issues/4121


johnhaley81:
Confirmed working. Thanks for all involved with finding this out!

PR #856, opened 2016-01-06
Extend stageFilemode to accept array of paths

kenprice:
...as well as just a string.

There is a case in GK where `Repository.prototype.stageFilemode` gets called a bunch of times (once for every file in working directory, for instance). This wastes resources. `stageFilemode` should be able to accept a single path as string or multiple paths as array, then stage/unstage filemode changes for files specified as usual.


johnhaley81:
LGTM. Thanks @kenprice

johnhaley81:
Scratch that... we need tests for it first.

PR #647, opened 2015-07-21
Update libgit2 to latest version

implausible:
# Update to Libgit2 0.23.0
### AnnotatedCommit
__NodeGit.AnnotatedCommit.fromRevspec(repository, revspec)__
NodeGit now allows finding commits by their extended sha syntax string.

### Branch
__NodeGit.Branch.create(repo, branchName, target, force)__
No longer has signature / log message.

__NodeGit.Branch.createFromAnnotated(repository, branchName, commit, force)__
Creates a branch at the specified commit. Force is optional.

__NodeGit.Branch.move(branch, newBranchName, force)__
No longer has signature / log message.

### Checkout
__CheckoutOptions__
CheckoutStragey no longer takes _NodeGit.Checkout.STRATEGY.SAFE_CREATE_. Use _NodeGit.Checkout.STRATEGY.SAFE_ | _NodeGit.Checkout.STRATEGY.RECREATE_MISSING_ in its place.

### Clone
__CloneOptions__
Clone Options no longer contains a RemoteCallbacks field, and instead contains a FetchOptions field which contains a RemoteCallbacks field.
This affects the any and all clone operations which were using RemoteCallbacks.

### Commit
__Commit.prototype.headerField(field)__
Returns a field from the headers of a commit.

### Config
__Config.prototype.getString(name)__
Config.prototype.getString() is deprecated and is no longer in the API. Use Config.prototype.getStringBuf(name) instead.

### Cred
__NodeGit.Cred.sshKeyMemoryNew(username, publickey, privatekey, passphrase)__
Returns a credentials object for the given username, publickey, privatekey, passphrase which are all strings.

### Diff
__NodeGit.Diff.blobToBuffer(buffer_as_path, options, file_cb, binary_cb, hunk_cb, line_cb, payload)__
Diff.blobToBuffer now takes an additional argument binary_cb which is the binary callback. Binary callback will be called for binary diffs.

### FetchOptions
__FetchOptions__
FetchOptions now contains a RemoteCallbacks field called callbacks.
Fields:
* __callbacks__ : _RemoteCallbacks_
* __prune__ - _NodeGit.Fetch.PRUNE_
	* _GIT_FETCH_PRUNE_UNSPECIFIED_
	* _GIT_FETCH_PRUNE_
	* _GIT_FETCH_NO_PRUNE_
* __updateFetchhead__ : _Number_ - defaults to 1, leave 1 in order to behave like git, treat 0 as off
[...]

PR #860, opened 2016-01-11
Fix clone options in two factor authorization

kamimoo:
Related to https://github.com/nodegit/nodegit/pull/647#issuecomment-142915106 and #848

The fixes similar to #848 is applied to cloning with two factor authorization as well.
And if you use `fetchOpts` you don't need a URL with a token any more.


johnhaley81:
This is great. Thank you for the PR @kamimoo

PR #862, opened 2016-01-14
Make submodule init / setup functions async

srajko:
This makes a number of NodeGit functions related to adding / opening / initializing a submodule async.  This is to better process the return value (getting a promise rejection with an error message instead of a numeric error code), and to reflect that the methods are affecting the repo in the file system.

Also added a test which exercises most of the affected functions (except for `update` - I am getting crashes on `update` that I still need to dig into, but the new NodeGit signature of the function should be correct).

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -1967,6 +1967,24 @@
     },
     "submodule": {
       "functions": {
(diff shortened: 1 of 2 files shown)


johnhaley81:
LGTM. Thanks @srajko.

PR #864, opened 2016-01-15
Add rudimentary diagnostics and testing for thread safety code

srajko:
Exposes a few thread safety internals so they can be at least somewhat tested.

Also adds support for an environment variable `NODEGIT_TEST_THREADSAFETY`, which when set will enable thread safety for the run of the test suite.


johnhaley81:
Thanks for the test cases @srajko.

PR #865, opened 2016-01-15
Configure travis to not use coverage and gcc4.9 during tagged releases

tbranyen:
Is it really this easy?

*... probably not!*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,14 +38,18 @@ before_install:
 
   - nvm install $NODE_VERSION
 
-  - if [ $TRAVIS_OS_NAME == "linux" ]; then
+  - if [ $TRAVIS_OS_NAME == "linux" ] && [ ! "$TRAVIS_TAG" ]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -38,14 +38,18 @@ before_install:
...
   - nvm install $NODE_VERSION
 
-  - if [ $TRAVIS_OS_NAME == "linux" ]; then
+  - if [ $TRAVIS_OS_NAME == "linux" ] && [ ! "$TRAVIS_TAG" ]; then
       export GYP_DEFINES="use_obsolete_asm=true";
       export JOBS=4;
       export CC=/usr/bin/gcc-4.9;
       export CXX=/usr/bin/g++-4.9;
+    else
+      export GYP_DEFINES="use_obsolete_asm=true";
```
johnhaley81: We default to that in https://github.com/nodegit/nodegit/blob/master/vendor/openssl/openssl.gyp#L12. That _might_ not be needed anymore?? /cc @implausible

PR #867, opened 2016-01-18
Submodule get set

srajko:
Adds support for `Submodule.status`, and makes a number of setters async to get better error handling, and match the behavior of the setters (writing to disk).


johnhaley81:
LGTM. Thanks @srajko.

Issue #869, opened 2016-01-19
Crash cleaning up mutexes


joshaber:
From https://github.com/atom/atom/pull/10445#issuecomment-172964977:

```
abort() called

Thread 0 Crashed:: CrRendererMain  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib          0x00007fff8de9d286 __pthread_kill + 10
1   libsystem_c.dylib               0x00007fff8b7da9b3 abort + 129
2   libnode.dylib                   0x000000010553da54 uv_mutex_destroy + 20
3   nodegit.node                    0x000000011002dd80 LockMasterImpl::CleanupMutexes(uv_async_s*) + 72
4   libnode.dylib                   0x000000010553461b 0x1053f4000 + 1312283

...

Thread 12:
0   libsystem_c.dylib               0x00007fff8b7d79ef strncasecmp_l + 84
1   nodegit.node                    0x00000001101cd763 git_index_entry_isrch + 47
2   nodegit.node                    0x000000011021476a git__bsearch + 67
3   nodegit.node                    0x00000001101d2708 git_index_snapshot_find + 73
4   nodegit.node                    0x00000001101d64fe workdir_iterator__enter_dir + 441
5   nodegit.node                    0x00000001101d70fa fs_iterator__expand_dir + 1050
6   nodegit.node                    0x00000001101d5e1b fs_iterator__advance_into + 67
7   nodegit.node                    0x00000001101bc073 git_diff__from_iterators + 2156
8   nodegit.node                    0x00000001101bcfa3 git_diff_index_to_workdir + 219
9   nodegit.node                    0x000000011020d79c git_submodule__status + 754
10  nodegit.node                    0x00000001101bc53a git_diff__from_iterators + 3379
11  nodegit.node                    0x00000001101bcfa3 git_diff_index_to_workdir + 219
12  nodegit.node                    0x000000011020973a git_status_list_new + 930
13  nodegit.node                    0x000000011016df9f GitStatusList::CreateWorker::Execute() + 67
14  libnode.dylib                   0x0000000105532d2e 0x1053f4000 + 1305902

...

Thread 13:
0   libsystem_kernel.dylib          0x00007fff8de9d14e __psynch_mutexdrop + 10
1   libnode.dylib                   0x000000010553dab9 uv_mutex_unlock + 9
2   nodegit.node                    0x000000011002deb5 LockMasterImpl::GetMutexes(int) + 227
[...]

PR #871, opened 2016-01-20
Fix unlocking of alreadyLocked mutex

srajko:
This is a somewhat conjectural attempt at a fix for https://github.com/nodegit/nodegit/issues/869

Someone please double-check my logic from that thread :-D


srajko:
https://github.com/nodegit/nodegit/issues/869#issuecomment-173110930

PR #872, opened 2016-01-21
Add support for Remote.defaultBranch

srajko:


test/tests/remote.js
```diff
@@ -150,6 +150,22 @@ describe("Remote", function() {
...
   });
 
+  it("can get the default branch of a remote", function() {
+    var remoteCallbacks = {
+      certificateCheck: function() {
+        return 1;
+      }
+    };
+
+    var remote = this.remote;
```
tbranyen: Can't wait until we can use `=>` syntax and avoid re-assignment like this.

johnhaley81:
:+1:

PR #878, opened 2016-01-22
Add missing 'promise' dependency

lowellk:


johnhaley81:
All of the versions of node that we target have `Promise` natively. Where is this breaking for you?

lowellk:
@johnhaley81 I'm running node 0.10. Is that not supported?

johnhaley81:
Unfortunately no. We decided to stop support for node v0.10 when 5 was released. Are you able to use NodeGit in v0.10 with this patch?

lowellk:
My system was running 0.10 but I realize the best solution for me is to simply upgrade my version. Thanks for responding.

PR #880, opened 2016-01-27
Clean mutexes as part of garbage collection cycle instead of uv_async

### LINKED ISSUE #866: Process not exiting by default
Running a js file that has has just one line that requires nodegit like so:
```javascript
var nodegit = require("nodegit");
```
results in the process not exiting... Is this a bug?
node in verson 4.2.4, nodegit in latest version from npm: 0.8.0

srajko:
Leaving the `uv_async` handle open was causing the node process to hang on exit.  Cleaning up mutexes on garbage collection achieves the same goal (mutexes will get cleaned up at some point, when they are no longer needed), and avoids the issue.

Using the GC is probably better performance-wise with thread safety enabled (mutex cleanup should be less frequent).  With thread-safety disabled, the mutex cleanup gets skipped (though for simplicity, the GC callback is configured throughout - I imagine this has minimal performance impact).

Fixes #866.

### DIFF
--- a/generate/templates/manual/src/lock_master.cc
+++ b/generate/templates/manual/src/lock_master.cc
@@ -1,3 +1,4 @@
+#include <nan.h>
 #include <git2.h>
 #include <uv.h>
 #include <set>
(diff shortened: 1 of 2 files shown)


tbranyen:
:heart_eyes:

Issue #877, opened 2016-01-22
calculating convenient patches locks files without autocrlf


implausible:
When calculating patches for a diff of the working directory, if autocrlf is unset or false, the libgit2 creates a file lock for all patches until the patches are freed.

Unfortunately, @srajko and I were working on this issue today and uncovered the fact that we are also not freeing objects that libgit 2 creates for us that libgit2 wants us to free. This means that when patches are created, even after the reference to the javascript object is GCd, the patch and associated handle exist out in heap land.

Due to this problem, it is necessary that we specify in our descriptor objects that need to be freed in the  GC process. I have a branch https://github.com/nodegit/nodegit/tree/free-raw-on-demand which begins the implementation of that. The branch also starts introducing the ability to free raw objects while in javascript.

The mechanisms for freeing the resources would also need to be templated out such that if we declare in javascript, free this object, the object should now return 'not allocated' errors or we should specify some other means of emptying the object.

Interestingly enough, since this error affects patches, which are directly linked to the hunks of the diff, I am also proposing that we move the convenience classes for patches, lines, and hunks into a manual C++ template. The reasoning behind this is that at the time that patches are calculated, it is possible to transfer all of that data into a single object that is not locking any files via libgit2. This convenience wrapper would discard all libgit2 references and purely represent the diff data. This means that our convenience class can serve that data to javascript land faster and more conveniently.

Looking for opinions.

srajko:
> Interestingly enough, since this error affects patches, which are directly linked to the hunks of the diff, I am also proposing that we move the convenience classes for patches, lines, and hunks into a manual C++ template.

I like the idea of prefetching data at the C++ level and making it available in ready-to-go JS format - that data would be accessible to the JS thread synchronously and at no risk of blocking with the new thread safety code (CC @johnhaley81 @joshaber @maxkorp @tbranyen).  It would be great to make that available for any data/type, but that's probably a longer/separate discussion.

The most critical part here is probably the other stuff - making sure we do GC those libgit2 objects that are currently getting leaked, and allowing for types that hold on to file handles to be freed explicitly.

johnhaley81:
[...]

PR #883, opened 2016-01-27
Better convenient patches

implausible:
This is the solution I've come up with to fix the locking issues with patches. See https://github.com/nodegit/nodegit/issues/877.


implausible:
This iteration is currently too slow, need to build a refcounter on hunks / lines and implement convenient line.

implausible:
Having difficulties tracking down where exactly the content pointer exists in libgit2 (the pointer that lines keep to the diff text). Instead, since this is a convenience wrapper, I am going to remove that pointer to the diff text in favor of the actual content of the line.

generate/templates/templates/nodegit.js
```diff
@@ -73,9 +81,9 @@ require("./utils/normalize_options");
 require("./utils/shallow_clone");
 
 // Load up extra types;
-require("./convenient_line");
-require("./convenient_hunk");
-require("./convenient_patch");
+// require("./convenient_line");
```
johnhaley81: Remove these please.

johnhaley81:
This looks great! Could we get a summary of the API changes involved in this?

implausible:
Minor changes to API:

- ConvenientPatch does not have a patch or a delta property associated with it, if you were using the delta, please just use prototype methods oldFIle, newFile, and Status, which are stripped directly from the delta.
- ConvenientPatch.prototype.hunks returns a promise with an array of ConvenientHunks.
- ConvenientPatch 
- ConvenientHunk does not have an exposed diffHunk associated with it, but does have the same members as diffHunk:
   - size() : number of lines in the hunk
   - oldStart() : old starting position
   - oldLines() : number of lines in old file
   - newStart() : new starting position
   - newLines() : number of lines in new file
   - headerLen() : length of header
   - header() : returns the header of the hunk
   - lines() : returns a promise containing DiffLines, not ConvenientLines.
- DiffLine now contains the members rawContent() and content().
   - rawContent() contains the unformatted content of the line. This is no longer a string from the line to the end of the file.
   - content() contains the utf8 formatted content of the line.

johnhaley81:
:star:

implausible:
@tbranyen thoughts before I merge this in?

tbranyen:
@implausible how's the performance looking now? Just did a brief scan of the code and the change looks fine to me, I think the move to C++ might make the convenient api's a bit harder to maintain.

implausible:
It will make it a bit harder to follow for sure. The performance is comparable to before. Unsure if faster or slower because it's pretty close. One thing that is nice is that this completely removes the locking issues we had with the git_patch_from_diff issue.

johnhaley81:
@implausible there was no perf increase?

implausible:
@johnh: Unsure if there is a perf 'increase', but there is a certain stability increase. There is definitely a positive perf gain over alternative systems to fix this issue. I can run some tests vs the old system. What i know for sure is there is 0 perf decrease. I think this will be as fast as the system can transfer data between git_patch and git_hunks and git_lines. It still runs asynchronously while the system does transfer of data.

tbranyen:
:+1:

johnhaley81:
I'm @johnhaley81. You just pinged some random guy. Sorry @johnh.

johnhaley81:
Could we get a delta before we merge? I'd like to include that in the changelog.

implausible:
slack habits lol

tbranyen:
If none of the interfaces have changed, we could probably just mention that they are more stable now and moved to C++

johnhaley81:
:+1: I'm good with this. @tbranyen I'm going to do a release to 0.10.0 after this is merged.

implausible:
Wow. I am wrong. There is a HUGE perf improvement. @tbranyen @johnhaley81 @srajko

Result before move to C++ convenience wrapper:
Time to generate patches: 254
number of patches: 927
Time to generate hunks: 34
Number of hunks: 890
Time to generate lines: 3465
Number of lines: 105705


Result after move:
Time to generate patches: 417
number of patches: 927
Time to generate hunks: 33
Number of hunks: 890
Time to generate lines: 127
Number of lines: 105705

johnhaley81:
:star2: :100: :star2:

johnhaley81:
Time to calc patches is up but that's because we're copying a lot of stuff into memory from the old patch structure. Overall an order of magnitude perf increase. Nice job!

PR #889, opened 2016-02-02
Implement file history walk in revwalk

implausible:
Implementing file history walk for revwalk. You pass it a sha, it then traverses until it either finds the 'added' entry for the file OR it hits the max number of commits to search for history.

### DIFF
--- a/generate/templates/manual/revwalk/file_history_walk.cc
+++ b/generate/templates/manual/revwalk/file_history_walk.cc
@@ -19,7 +19,7 @@ NAN_METHOD(GitRevwalk::FileHistoryWalk)
   String::Utf8Value from_js_file_path(info[0]->ToString());
   baton->file_path = strdup(*from_js_file_path);
   baton->max_count = (unsigned int)info[1]->ToNumber()->Value();
-  baton->out = new std::vector< std::pair<git_commit*, char *> *>;
+  baton->out = new std::vector< std::pair<git_commit *, std::pair<char *, git_delta_t> > *>;
   baton->out->reserve(baton->max_count);
   baton->walk = Nan::ObjectWrap::Unwrap<GitRevwalk>(info.This())->GetValue();
 
@@ -36,8 +36,6 @@ void GitRevwalk::FileHistoryWalkWorker::Execute()
   git_repository *repo = git_revwalk_repository(baton->walk);
   git_oid *nextOid = (git_oid *)malloc(sizeof(git_oid));
   giterr_clear();
-  char * trackPath = strdup(baton->file_path);
-  bool addedFlag = false;
   for (
     unsigned int i = 0;
     i < baton->max_count && (baton->error_code = git_revwalk_next(nextOid, baton->walk)) == GIT_OK;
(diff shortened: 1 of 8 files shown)


johnhaley81:
https://github.com/nodegit/nodegit/blob/master/examples/walk-history-for-file.js needs to be updated as well please.

implausible:
@tbranyen @johnhaley81 once we get this to pass without some random test timing out, this is ready to go by my account. If you guys wanted to give this the once over before I merge in, that would be greatly appreciated.

Notes of interest: The fileHistoryWalk will do a rename detection if it discovers a file was added in the history and will return the old file name and the renamed status along with the commit. This helps history users to not only follow the filepath's history, but also see how the content if the file has changed / branched.

generate/templates/manual/revwalk/file_history_walk.cc
```diff
@@ -0,0 +1,270 @@
...
+      break;
+    }
+
+    git_tree *thisTree, *parentTree;
+    if ((baton->error_code = git_commit_tree(&thisTree, nextCommit)) != GIT_OK) {
+      git_commit_free(nextCommit);
+      break;
+    }
+
+    git_diff *diffs;
```
johnhaley81: I don't see these getting freed anywhere. Is this memory cleaned up?

generate/templates/manual/revwalk/file_history_walk.cc
```diff
@@ -0,0 +1,270 @@
...
+
+      if (baton->error_code < GIT_OK) {
+        break;
+      }
+
+      if (nextPatch == NULL) {
+        continue;
+      }
+
+      const git_diff_delta *delta = git_patch_get_delta(nextPatch);
```
johnhaley81: Just a quick note that we do *not* need to free this https://libgit2.github.com/libgit2/#v0.23.4/group/patch/git_patch_get_delta

johnhaley81:
Other then the potential memory leak above, could you get some comment blocks into `lib/revwalk/js` for docs?

generate/templates/manual/revwalk/file_history_walk.cc
implausible: I can't find whether or not we're supposed to clean it up or not. It would be super helpful if the docs gave me some insight

implausible: If you know explicitly that we're supposed to clean those up, I'll free the. I've been mulling it over myself, because the docs are not clear!

johnhaley81: Yes we do need to free it: https://github.com/libgit2/libgit2/blob/v0.23.4/include/git2/diff.h#L23-L28

implausible: oh cool! I guess since it's explained there they don't have to explain it in the function documentation for tree to tree.

generate/templates/manual/revwalk/file_history_walk.cc
implausible: Yea, thanks for linking this in case anyone didn't know. the delta is attached to the patch object. When we free the patch it is freed, as well.

generate/templates/manual/revwalk/file_history_walk.cc
```diff
@@ -0,0 +1,266 @@
...
+
+        git_patch_free(nextPatch);
+
+        if (flag) {
+          break;
+        }
+      }
+    }
+
+    git_diff_free(diffs);
```
implausible: Freeing diffs down here now.

implausible:
@tbranyen what is going on with this test? https://travis-ci.org/nodegit/nodegit/jobs/106996362

PR #891, opened 2016-02-03
make checkoutBranch works with branches that lives outside of refs/heads

AsoSunag:
And avoid trying to call a method on a boolean when isBranch() is false.

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -1146,9 +1146,6 @@ Repository.prototype.checkoutBranch = function(branch, opts) {
       NodeGit.Checkout.STRATEGY.RECREATE_MISSING);
   return repo.getReference(branch)
   .then(function(ref) {
-    if (!ref.isBranch()) {
-      return false;
-    }
     reference = ref;
     return repo.getBranchCommit(ref.name());
   })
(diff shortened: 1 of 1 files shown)


johnhaley81:
So I could be wrong but I'm pretty sure that anything that exists outside of `refs/heads` is, according to git, *not* a branch. So this function is doing what it is supposed to be doing. I definitely see the use case for this though.

Let's abstract this out to `checkoutRef` and then make another method that's called `checkoutBranch` that has the old functionality. Maybe have `checkoutBranch` confirm that the ref is indeed a branch before calling `checkoutRef` and then rejecting if it isn't.

AsoSunag:
Hello,

Yes you're right, I did that for remotes. It is good idea to have a checkoutRef and a checkoutBranch. The only thing that is bothering me with that is that when using checkoutRef with a non branch reference we will still use getBranchCommit. So we could keep one method checkoutBranch checking if it is a branch or a remote or we can have two methods.

What do you think is the best?

johnhaley81:
`Repository#getBranchCommit` just calls `Repository#getReferenceCommit` so we can replace that call with `Repository#getReferenceCommit`

AsoSunag:
ah ok I'll do that then.

johnhaley81:
LGTM Thanks @AsoSunag

PR #893, opened 2016-02-04
Write docs for convenient patch and convenient hunk

implausible:


johnhaley81:
@implausible linter errors

PR #894, opened 2016-02-04
Bump timeout for tests to 15 seconds and remove coverage from release builds

johnhaley81:
We see a lot of test failures due to network latency and this should help that.


johnhaley81:
The best is when you're doing a release and the last platform on appveyor dies due to a timeout and you have to restart the entire 30-minute process all over again.

PR #896, opened 2016-02-05
Hot fix historywalk

implausible:
was dropping deleted events due to rename detection. oops.


johnhaley81:
Let's confirm that things are fixed 100% before we pull this in and do another release.

implausible:
@johnhaley81 Added test to confirm both directions of rename detection work now.

implausible:
This will pass, am done with file history now.

PR #899, opened 2016-02-08
Fix documentation issues

jvilk:
Fixes a variety of documentation issues:

* Fixes cases where the docs said `Ref`, but meant `Reference`.
* Adds documentation for `DiffFile`.
  * NOTE: There's a bug in your doc generator, which causes `DiffFile` to display the invalid instance fields alongside my correct methods.
* Fixes `exports.module` => `module.exports` in a few places.

Let me know if you want me to squash these into one commit.


tbranyen:
:+1: Thanks!

PR #906, opened 2016-02-16
Fix stageLines

joshaber:
There were a couple funny things:

1. Removing the index lock seems out of place for this function.
2. We should calculate the diff *after* refreshing the index.

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -1361,20 +1361,22 @@ Repository.prototype.stageLines =
 
   var repo = this;
   var index;
-  var diffPromise = isSelectionStaged ?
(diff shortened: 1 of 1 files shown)


lib/repository.js
```diff
@@ -1409,18 +1411,13 @@ Repository.prototype.stageLines =
       });
     };
 
-  var indexLock = repo.path().replace(".git/", "") + ".git/index.lock";
-
-  return fse.remove(indexLock)
```
johnhaley81: :+1:

johnhaley81:
LGTM. Gracias amigo.

PR #908, opened 2016-02-18
Don't check for \\ No newline at end of file\n if it can't exist

implausible:
If content isn't greater than 29, don't search for the no newline token. This lead to segfaults on linux. Interestingly, this did not affect most OSX and Windows machines. Some sort of compiler-saving-of-the-day?? No clue.


tbranyen:
Magic number? Would it be possible to document that somewhere in the code?

johnhaley81:
@implausible is it 29 because that's the length of [this string](https://github.com/nodegit/nodegit/pull/908/files#diff-d82e95bb7d148558bc2ea99c37aae6ddR85)?

tbranyen:
That makes way more sense, at least we know it's not magic.

PR #909, opened 2016-02-18
Fix `fast_walk.cc` when there was an error

johnhaley81:
In the error state, fast walk would just do an equality check instead of an assignment of the error. I'm an idiot :(


tbranyen:
:lollipop:

Issue #907, opened 2016-02-17
Malloc bug with fetchAll and ssh.


holblin:
Hi,

I have a bug with fetchAll action.

I make this simple code for testing this bugs:

```javascript
'use strict';
var path = require('path');
var Git = require('nodegit');

var fetchOpts = {
  callbacks: {
    certificateCheck: function() { return 1; },
    credentials: function(url, userName) {
      return Git.Cred.sshKeyNew('git', path.resolve('/Users/holblin/.ssh/id_rsa.pub'), path.resolve('/Users/holblin/.ssh/id_rsa'), '');
    }
  }
};

Promise.resolve()
.then(function() {
  var all =[];
  for (var i = 0; i < 2; i++) {
    all.push(updateIt('repo_'+i));
  }
  return Promise.all(all);
})

function updateIt(i) {
  var repository, localHeadCommit;
  var dir = 'tests/'+i;

  return Git.Repository.open(dir)
    .then((repo) => repository = repo) // save the reference for futur use
    .then(() => console.log('['+i+'] openRepo'))
    .then(() => repository.fetchAll(fetchOpts)) // fetch all data from all remotes
    .then(() => console.log('['+i+'] fetchAll'))
}
```

For use it, make this commands first:
```bash
npm -i path nodegit
mkdir tests
cd tests
git clone git@<repository1_git_url>.git repo_0
git clone git@<repository2_git_url>.git repo_1
cd ..
```

I have a Mac os, X yosemite version 10.10.5. I use nvm with node v5.5.0.
My uname -a is `Darwin zolesio 14.5.0 Darwin Kernel Version 14.5.0: Tue Sep  1 21:23:09 PDT 2015; root:xnu-2782.50.1~1/RELEASE_X86_64 x86_64`.
The nodegit version is: `nodegit@0.11.1`

Sometimes, it finished with:
```text
[repo_0] openRepo
[repo_1] openRepo
node(5166,0x1026cd000) malloc: *** error for object 0x102303d00: pointer being realloc'd was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

Sometimes with:
```text
[repo_0] openRepo
[repo_1] openRepo
node(5058,0x105087000) malloc: *** error for object 0x10170e960: double free
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

Sometimes with:
```text
[repo_0] openRepo
[repo_1] openRepo
Bus error: 10
```

[...]

PR #910, opened 2016-02-19
[WIP] Chain async actions in promises to prevent overlap

srajko:
A possible fix for #907, opening PR to run tests.  If it works, it will merit a longer explanation / discussion.

Credit for the beautiful idea of using JS promises to prevent any overlap in async libgit2 calls goes to @johnhaley81.

### DIFF
--- a/generate/templates/templates/nodegit.js
+++ b/generate/templates/templates/nodegit.js
@@ -3,8 +3,9 @@ var promisifyNode = require("promisify-node");
 var asyncPromise = Promise.resolve();
 
 function promisify(fn) {
+  var promisified = promisifyNode(fn);
+
   function chainedPromisified() {
-    var promisified = promisifyNode(fn);
     var _this = this;
     var _args = arguments;
 
--- a/generate/templates/templates/nodegit.js
+++ b/generate/templates/templates/nodegit.js
@@ -1,4 +1,30 @@
-var promisify = require("promisify-node");
+var promisifyNode = require("promisify-node");
+
+var asyncPromise = Promise.resolve();
+
+function promisify(fn) {
+  function chainedPromisified() {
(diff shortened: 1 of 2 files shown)


generate/templates/templates/nodegit.js
```diff
@@ -1,4 +1,30 @@
...
+      function runPromisified() {
+        return promisified.apply(_this, _args)
+          .then(function(value) {
+            resolve(value);
+          }, function(error) {
+            reject(error);
+          });
+      }
+      asyncPromise = asyncPromise
+        .then(runPromisified);
```
johnhaley81: This should probably be `then(runPromisifed, runPromisified)`

johnhaley81:
I think we should also include an `nonBlocking` option for each `async` call. That way the consumer of the library can decide whether or not they want something to run in a "blocking" state or not.

Ex: Say I want to do a `fetch` but I also want to calculate the `diff` of a commit at the same time. These don't really affect each other and the `fetch` can take a very long time. It would suck to have to wait each time for the `fetch` when I could just handle that case.

I would say that `nonBlocking` would be the exception though since you almost always want the safety without having to think about it.

generate/templates/templates/nodegit.js
```diff
@@ -1,4 +1,30 @@
-var promisify = require("promisify-node");
+var promisifyNode = require("promisify-node");
+
+var asyncPromise = Promise.resolve();
+
+function promisify(fn) {
+  function chainedPromisified() {
+    var promisified = promisifyNode(fn);
```
johnhaley81: This should be up in the previous scope so we don't promisify the function every time it's called.

johnhaley81:
This is no longer needed.

PR #912, opened 2016-02-19
OpenSSL and libssh2 thread safety

srajko:
Provide openssl with the locking setup it needs for thread safety.

Initialize libssh2 to avoid the possibility of its subsequent thread-unsafe init.

### DIFF
--- a/generate/templates/manual/include/init_ssh2.h
+++ b/generate/templates/manual/include/init_ssh2.h
@@ -1,0 +1,6 @@
+#ifndef INIT_SSH2
+#define INIT_SSH2
+
+void init_ssh2();
+
+#endif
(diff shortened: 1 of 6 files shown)


srajko:
The `libssh2_init` fix resolves the crash from #907.  Calls to fetch would initialize openssh sessions in separate threads, and those would in turn call `libssh2_init` via https://github.com/libssh2/libssh2/blob/d441da3086dcc49989b7c5b9bf3043c4a6a98211/src/global.c#L74-L78 which is not thread-safe.

Providing openssl with the callbacks it needs to operate in a thread safe manner also seems to help, and is probably something we should be doing anyway.

srajko:
There was a snag with doing the libssh2 initialization directly from `nodegit.cc` because `node.h` and `libssh2.h` had conflicting type definitions.  The final fix for the problem was to move the `libssh2.h` include to a separate file and do the initialization from there.

johnhaley81:
LGTM

PR #918, opened 2016-02-24
Fix debug build on windows

johnhaley81:
There was a linking exception that was fixed by changing the order of includes for node and v8 headers

### DIFF
--- a/generate/templates/templates/nodegit.cc
+++ b/generate/templates/templates/nodegit.cc
@@ -1,7 +1,7 @@
 // This is a generated file, modify: generate/templates/nodegit.cc.
+#include <node.h>
 #include <v8.h>
 
-#include <node.h>
 #include <git2.h>
 #include <map>
 #include <algorithm>


generate/templates/templates/nodegit.cc
```diff
@@ -1,7 +1,7 @@
 // This is a generated file, modify: generate/templates/nodegit.cc.
```
joshaber: :trollface:?

johnhaley81: That's there in case somebody is looking at the generated file `src/nodegit.cc` and is like "why can't I commit my changes.

maxkorp:
Yeah, i wish we had a good way to point to the root template in combyne so we didnt have to include that in the _not_ generated files.

generate/templates/templates/nodegit.cc
tbranyen: Should we add a Combyne comment above it?

``` html
{%-- JavaScript comment below alerts contributors to not edit the generated source --%}
```

maxkorp: Probably. That would at least clarify a bit. Might be nice if we just do that when we get the compiled output back, like 
`// This is a generated file, modify: {root template}.` in the `generateNativeCode.js`

maxkorp:
https://github.com/nodegit/nodegit/pull/924

johnhaley81:
:+1:

PR #923, opened 2016-02-25
Remove `GetRefValue`

johnhaley81:
Using `GetRefValue` would return the V8 handled pointer instead of the pointer to the libgit2 value and the V8 handled pointer could be changed at any moment.

Removing this function and setting the pointer value to the underlying C object will prevent V8 GC screwing up a NodeGit app.


johnhaley81:
@joshaber I want to get this in before another release. This is pretty big.

joshaber:
That's fine, I'm in no rush.

PR #927, opened 2016-02-29
fixes #852 should not be an array with null in it

### LINKED ISSUE #852: Should Repository#createCommitOnHead create a new head if tree is empty?
Using this function on a repository with no commits currently yields with "Error: Oid id is required." The function is quite helpful but i have to write the long way if the there is no HEAD. Which makes the shorter version irrelevant once i took the effort to write the longer one. So should the option be included to also support the initial commit if none is available?

get-trucked:
I'm not sure if this is the intended behaviour, but my take on it as a convenience method it should be able to handle creating the initial commit on a empty repo.

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -535,9 +535,9 @@ Repository.prototype.createCommitOnHead = function(
     .then(function(treeOid) {
       return repo.getHeadCommit()
         .then(function(parent) {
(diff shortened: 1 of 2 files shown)


johnhaley81:
:+1: Thanks @get-trucked

PR #930, opened 2016-02-29
Let my patches go

### LINKED ISSUE #877: calculating convenient patches locks files without autocrlf

joshaber:
Fixes #877.

It looks like #883 was supposed to fix this, but we missed the part where we actually did the free’ing. :see_no_evil: 

This is only WIP because I’m still trying to make a build so I can test that this fixes the whole problem.

### DIFF
--- a/generate/templates/manual/patches/convenient_patches.cc
+++ b/generate/templates/manual/patches/convenient_patches.cc
@@ -62,6 +62,7 @@ void GitPatch::ConvenientFromDiffWorker::Execute() {
 
       if (nextPatch != NULL) {
         baton->out->push_back(createFromRaw(nextPatch));
+        patchesToBeFreed.push_back(nextPatch);
       }
     }
 
--- a/generate/templates/manual/src/convenient_patch.cc
+++ b/generate/templates/manual/src/convenient_patch.cc
@@ -68,7 +68,7 @@ PatchData *createFromRaw(git_patch *raw) {
     hunkData->lines = new std::vector<git_diff_line *>;
     hunkData->lines->reserve(hunkData->numLines);
 
-
+    static const int noNewlineStringLength = 29;
     bool EOFFlag = false;
     for (unsigned int j = 0; j < hunkData->numLines; ++j) {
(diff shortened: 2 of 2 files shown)


joshaber:
(Also cc https://github.com/atom/atom/issues/10931.)

johnhaley81:
Yeah, I swore I commented on that exact line saying "These need to be freed" but I can't find it now. /shrug

implausible:
Oh wow, good catch. Thanks!

implausible:
That should take care of the problem, the intention of the system as it was before was to do exactly that. I just accidentally a line.

joshaber:
Verified this fixes the issue we were seeing in Atom :+1:

johnhaley81:
:+1: Thanks @joshaber

PR #931, opened 2016-03-01
Bump `http-parser` to v2.6.1

### LINKED ISSUE #911: 5.6.0 https url:  [Error: HTTP parser error: unexpected content-length header]
With node 5.6.0 which fixes https parsing errors and changes current parsing behavior (https://nodejs.org/en/blog/vulnerability/february-2016-security-releases) trying to git clone or rebase https bitbucket private url results in
```
[Error: HTTP parser error: unexpected content-length header]
```

standard clone example will reproduce this error
```
 nodegit.Clone(
// basic auth bitbucket repo is what I was testing
    "https://user:pass@bitbucket.org/repo.git",
    path,
    {
      fetchOpts: {
        callbacks: {
          certificateCheck: function() {
            // github will fail cert check on some OSX machines
            // this overrides that check
            return 1;
          }
        }
      }
    })
```

node 5.5.0 works as expected

this seems like an error from http-parser itself so I don't know if this is really related to nodegit, since I'm very short on time to do some more debugging

johnhaley81:
This should fix #911

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ env:
   matrix:
     - export NODE_VERSION="0.12"
     - export NODE_VERSION="4.1"
-    - export NODE_VERSION="5.0"
+    - export NODE_VERSION="5.6"
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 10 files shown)


markov00:
Hi, sorry but I still receive the same error of #911 [Error: HTTP parser error: unexpected content-length header] . I'm using nodegit 0.11.6 and node 5.6.0 in osx. Also tried with node 5.7

Issue #919, opened 2016-02-24
Transfer progress notifications are prohibitively slow


bpeabody:
Having even an empty transfer progress function for fetches adds a large amount of overhead. In the code below, for example, on the nodegit repo itself, specifying a transfer progress function adds about 38s for me.

The progress function is called once for each object fetched from the repository: about 25,000 times for the nodegit repo as of today. That would indicate an overhead of about 1.5ms for calling from C++ into JavaScript (running on a 3.1ghz i7). This amount of overhead feels a little high; I wonder if there might be some room for optimizing the JS/C++ binding/bridge code.

Even if significantly optimized (e.g., by a factor of 10), the overhead of calling from C++ into JS is probably high enough to be prohibitive; it might be good to be able to provide some conflation so that the progress function is called no more than a specified frequency (e.g., no more than once every 100ms).

If no one has an easy answer for this, I would be more than happy to lend a hand (I have a lot of experience with v8).

Sample code:

    const NodeGit = require("nodegit");

    console.time("no-progress");

    NodeGit.Clone("https://github.com/nodegit/nodegit.git", "no-progress")
    .then(function () {
        console.timeEnd("no-progress");
        console.time("progress");
        var options = new NodeGit.CloneOptions();
        var fetchOpts = new NodeGit.FetchOptions();
        options.fetchOpts = fetchOpts;
        var callbacks = new NodeGit.RemoteCallbacks();
        fetchOpts.callbacks = callbacks;
        callbacks.transferProgress = function () {
        };
        return NodeGit.Clone("https://github.com/nodegit/nodegit.git",
                             "progress",
                             options);
    }).done(function () {
        console.timeEnd("progress");
        process.exit(0);
    });

johnhaley81:
Wow, that's huge. You're totally right we need to throttle this at the C level. 

The code that would handle this is in [callback_helpers.cc](https://github.com/nodegit/nodegit/blob/master/generate/templates/partials/callback_helpers.cc#L4-L31) and [field_accessors.cc](https://github.com/nodegit/nodegit/blob/master/generate/templates/partials/field_accessors.cc#L85-L112).

Those are the static methods that we send to libgit2 to use as the proxy for the JS callbacks that will eventually be called. If we throttle these then this will work.

[...]

PR #932, opened 2016-03-01
Add throttling support for callbacks

srajko:


johnhaley81:
This will fix #919

johnhaley81:
@srajko can you add a test to confirm that the callback is indeed being throttled? It could be another clone test.

srajko:
The changes are as follows:

- Progress callbacks (`git_checkout_progress_cb`, `git_diff_file_cb`, `git_stash_apply_progress_cb`, `git_transfer_progress_cb`) now have default throttling of ~1 callback per 100ms
- All callbacks provided through configuration objects can now be throttled - instead of passing a function, pass an object `{ callback, throttle }` where `throttle` is the throttle period (~1 callback per throttle period)

A rundown of the commits:
 - `Add AsyncBatonWithResult and ExecuteAsync` - just refactoring, eliminating duplicated code
 - `Store defaultResult in baton` - I originally tried doing the throttling in the baton but that ended up being a mistake.  This moved the default result to the baton so the baton could use it.  I don't think the change is necessary for the final version of this PR, but I still like moving things out of combyne-templated files so I kept it.
 - `Refactor structs to use CallbackWrapper for callbacks` - just refactoring, leverages the previously unused `CallbackWrapper` to set us up for bundling throttling state with the javascript callback.
 - `Add callback throttle functionality` - the actual throttle
 - `Specify throttle in progress callback descriptors` - default settings for progress callbacks.

johnhaley81:
:+1: nice job! Thanks @srajko

generate/templates/manual/include/async_baton.h
```diff
@@ -13,4 +16,33 @@ struct AsyncBaton {
   bool done;
 };
 
+template<typename ResultT>
```
johnhaley81: :100:

PR #933, opened 2016-03-02
Test for Index.conflictAdd

dkoontz:


tbranyen:
:+1: always brightens my day when I see a PR with random tests of kindness

johnhaley81:
LGTM :+1:

PR #935, opened 2016-03-04
Docs: improved Merge.merge param docs

smith-kyle:
Clarify that their commit should be an `AnnotatedCommit` not a `Commit`


johnhaley81:
@smith-kyle thanks!

PR #936, opened 2016-03-04
Tests: generalized createConflict in index utils

smith-kyle:
The `createConflict` function in `test/utils/index_setup.js` was specialized for a specific index test. This PR generalizes it so it can be used in future tests.

PR #937, opened 2016-03-04
Expose mergehead_foreach

smith-kyle:
The test in this PR depends on #936


johnhaley81:
Thanks again @smith-kyle

PR #938, opened 2016-03-04
Fix diffing blobs to buffers with unicode

joshaber:
libgit2 wants the string’s byte count, while `String.length`gives us the number of characters. These aren’t necessarily the same thing in unicode land.

### DIFF
--- a/lib/diff.js
+++ b/lib/diff.js
@@ -68,7 +68,15 @@ Diff.blobToBuffer= function(
     binary_cb,
     hunk_cb,
     line_cb) {
-  var bufferLength = !buffer ? 0 : buffer.length;
+  var bufferText;
+  var bufferLength;
+  if (buffer instanceof Buffer) {
+    bufferText = buffer.toString("utf8");
+    bufferLength = buffer.length;
+  } else {
+    bufferText = buffer;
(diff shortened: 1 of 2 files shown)


joshaber:
This is ready now :+1:

lib/diff.js
```diff
@@ -68,7 +68,15 @@ Diff.blobToBuffer= function(
     binary_cb,
     hunk_cb,
     line_cb) {
-  var bufferLength = !buffer ? 0 : buffer.length;
+  var bufferText;
+  var bufferLength;
+  if (buffer instanceof Buffer) {
+    bufferText = buffer.toString("utf8");
```
tbranyen: Should we always assume utf-8? Wondering if we should make the encoding configurable with utf8 as a default?

joshaber: Yeah that's a good question. Certainly as a rule we shouldn't assume it. That said, we *do* assume it all over the place: e.g., https://github.com/nodegit/nodegit/blob/47e0d206a3b1a1f8114218aaee0fce514e9a9e98/lib/diff_line.js#L25 and anywhere that uses `toString` on a buffer.

I'd say addressing different encodings is a task for another day.

tbranyen: Makes sense, thanks @joshaber

PR #942, opened 2016-03-07
Revert "Merge pull request #931 from nodegit/bump-http-parser"

johnhaley81:
Bumping `http-parser` to 2.6.1 broke fetching/pulling from bitbucket in earlier versions of node somehow. This undoes that change and we'll have to look into fixing node v5.6.x later.


johnhaley81:
This reverts #931

PR #943, opened 2016-03-07
Revert "Merge pull request #932 from srajko/callback-throttle"

johnhaley81:
A non-trivial number of segfaults are seeing this in production. Pulling this back for right now until a later release where we can fix them.


johnhaley81:
Going to reopen #919 until this is fixed again without the double-freeing segfault.

PR #944, opened 2016-03-09
Throttle callbacks (again) + crash fix

srajko:
This reverts the removal of throttling, and adds two changes that seem to resolve the crash.  The likely fix is https://github.com/nodegit/nodegit/commit/1e585eae28fae98e1807cf20d88e643dae5cf7de, which restores some logic I omitted in the refactoring, thinking it should not make a difference... but it looks like it did.

I am a little worried about the state of the raw object making a difference after its wrapper has been destroyed, but maybe I'm just missing the big picture of what's happening here.  In any case, the logic matches what we were doing before, and no crash is better than crash.

### DIFF
--- a/generate/input/callbacks.json
+++ b/generate/input/callbacks.json
@@ -96,7 +96,8 @@
       "type": "int",
       "noResults": 1,
       "success": 0,
-      "error": -1
+      "error": -1,
+      "throttle": 100
     }
   },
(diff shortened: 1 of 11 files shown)


srajko:
This will fix #919

johnhaley81:
We've been running this in GK now with no reports of breakage. :+1: this is gtg. Thanks @srajko!

PR #945, opened 2016-03-09
Fix patch crash

joshaber:
[Libgit2 documents](https://github.com/libgit2/libgit2/blob/d8f721592817cb300339c827579bc545844b276f/include/git2/diff.h#L555) that `content` isn’t NUL-terminated, so we can’t
call `strlen` on it. If we do, we’ll probably crash.

### DIFF
--- a/generate/templates/manual/src/convenient_patch.cc
+++ b/generate/templates/manual/src/convenient_patch.cc
@@ -88,7 +88,7 @@ PatchData *createFromRaw(git_patch *raw) {
           calculatedContentLength > noNewlineStringLength &&
           !strncmp(
               &line->content[calculatedContentLength - noNewlineStringLength],
-              "\n\\ No newline at end of file\n", std::min(calculatedContentLength, noNewlineStringLength)
+              "\n\\ No newline at end of file\n", (std::min)(calculatedContentLength, noNewlineStringLength)
         )) {
           EOFFlag = true;
         }
--- a/generate/templates/manual/src/convenient_patch.cc
+++ b/generate/templates/manual/src/convenient_patch.cc
@@ -83,13 +83,12 @@ PatchData *createFromRaw(git_patch *raw) {
       }
 
       if (j == 0) {
-        // calculate strlen only once for the first line of the first hunk.
-        int calculatedContentLength = strlen(line->content);
+        int calculatedContentLength = line->content_len;
         if (
           calculatedContentLength > noNewlineStringLength &&
-          !strcmp(
+          !strncmp(
               &line->content[calculatedContentLength - noNewlineStringLength],
-              "\n\\ No newline at end of file\n"
+              "\n\\ No newline at end of file\n", std::min(calculatedContentLength, noNewlineStringLength)
         )) {
           EOFFlag = true;
         }
--- a/generate/templates/manual/src/convenient_patch.cc
+++ b/generate/templates/manual/src/convenient_patch.cc
@@ -56,8 +56,12 @@ PatchData *createFromRaw(git_patch *raw) {
 
   for (unsigned int i = 0; i < patch->numHunks; ++i) {
     HunkData *hunkData = new HunkData;
-    const git_diff_hunk *hunk;
-    git_patch_get_hunk(&hunk, &hunkData->numLines, raw, i);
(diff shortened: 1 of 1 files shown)


generate/templates/manual/src/convenient_patch.cc
```diff
@@ -72,17 +76,19 @@ PatchData *createFromRaw(git_patch *raw) {
...
-      const git_diff_line *line;
-      git_patch_get_line_in_hunk(&line, raw, i, j);
+      const git_diff_line *line = NULL;
+      int result = git_patch_get_line_in_hunk(&line, raw, i, j);
+      if (result != 0) {
+        continue;
+      }
 
       if (j == 0) {
-        // calculate strlen only once for the first line of the first hunk.
```
joshaber: This comment doesn't really describe what we were doing, and I'm not sure I understand what we're trying to do. So I'm not sure if this comment was wrong, or the code is wrong :sweat_smile:

joshaber:
This fixes a crash we saw in Atom. The crashlog would look like this:

```
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000125f72000

...

Thread 14 Crashed:
0   libsystem_c.dylib             	0x00007fff863db188 strlen + 72
1   nodegit.node                  	0x00000001164ef072 createFromRaw(git_patch*) + 525
2   nodegit.node                  	0x00000001165b7069 GitPatch::ConvenientFromDiffWorker::Execute() + 153
3   libnode.dylib                 	0x0000000110c4c48e 0x110b08000 + 1328270
4   libnode.dylib                 	0x0000000110c4c68b 0x110b08000 + 1328779
5   libnode.dylib                 	0x0000000110c57189 0x110b08000 + 1372553
6   libsystem_pthread.dylib       	0x00007fff968e9c13 _pthread_body + 131
7   libsystem_pthread.dylib       	0x00007fff968e9b90 _pthread_start + 168
8   libsystem_pthread.dylib       	0x00007fff968e7375 thread_start + 13
```

joshaber:
This is ready :pray:

johnhaley81:
@implausible could you also :eyes: this?

generate/templates/manual/src/convenient_patch.cc
implausible: This comment means that we don't want to calculate the length of content _every_ iteration of the loop because that is numLinesInContent * strlen iterations. This slows down performance substantially.

joshaber: I guess my confusion is that (1) it's not like we were re-using the length, and (2) we're doing this for the first line of every hunk, not just the first line of the first hunk as the comment says.

implausible: So the way that libgit2 represents the content pointer is like this:
Content ptr points to the start of the first line in the hunk, but continues until the end of the file.
content_len is the length of the line that we are looking at in the hunk
\n\\ No newline at end of file\n" is located at the end of content ptr (not at content_len).
content ptr continues to the last line of the file because content ptr does not terminate at the end of lines.

What we need to know at any given line is whether there is no new line at the end of the file as this is used in nodegit for staging. When I worked on this feature it was requested that we preserve the api as closely as possible. The best way to preserve the api was to append the no new line flag on the end of lines where we have discovered a file does not contain a new line at the end.

implausible: I'd like to investigate what a diff line looks like after this change before we commit this before we commit to this change.

implausible: No, you're right to be a bit confused. lol.

implausible: Yes, the comment is completely wrong.

implausible:
This looks good to me :+1:. I don't even know how I got to this solution anymore. So huge thanks to Josh for fixing yet another one of my errors. :heart:

implausible:
@joshaber took a quick walk. I think I understand where I got this wrong now. The no new line flag only matters for the last hunk in the diff. I assumed it was required for _every_ line of _every_ hunk. This is wrong. Thanks again!

joshaber:
Ah, I see. Thanks for clarifying @implausible!

PR #952, opened 2016-03-15
Fix Node 5.8.0

johnhaley81:


johnhaley81:
@tbranyen is your server down?

tbranyen:
@johnhaley81 yup, now is a good time for us to switch off it. Can you skip the test for now?

johnhaley81:
You got it. Thanks man!

PR #955, opened 2016-03-17
[WIP] Fix memory leak of repository and repo-cached objects

srajko:


implausible:
I've completed this in master.

PR #956, opened 2016-03-18
Stash.apply: Fix a typo in arguments to normalizeOptions

tylerchurch:
I tried to use `nodegit.Stash.apply` and it crashed... Looks like `NodeGit.CheckoutOptions` wasn't capitalized like it should've been.


johnhaley81:
Thanks for the fix @tylerchurch

PR #957, opened 2016-03-18
Use type safe pointer instead of void * for New

srajko:
Ran into this as part of https://github.com/nodegit/nodegit/pull/955.

Changing the `raw` pointer from `void *` to `const {{ cType }} *` to add type safety.  The change actually caught a problem in `struct_content.cc` where we were sometimes taking the address of a pointer and passing that in instead of the pointer (for `git_checkout_options.baseline`, we were sending in a `git_tree **` instead of a `git_tree *` to `New`).

### DIFF
--- a/generate/templates/templates/struct_content.cc
+++ b/generate/templates/templates/struct_content.cc
@@ -73,7 +73,7 @@ void {{ cppClassName }}::ConstructFields() {
       {% if not field.isEnum %}
         {% if field.hasConstructor |or field.isLibgitType %}
           Local<Object> {{ field.name }}Temp = {{ field.cppClassName }}::New(
-            &this->raw->{{ field.name }},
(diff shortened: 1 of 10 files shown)


generate/templates/templates/struct_content.cc
```diff
@@ -73,7 +73,7 @@ void {{ cppClassName }}::ConstructFields() {
       {% if not field.isEnum %}
         {% if field.hasConstructor |or field.isLibgitType %}
           Local<Object> {{ field.name }}Temp = {{ field.cppClassName }}::New(
-            &this->raw->{{ field.name }},
+            {%if not field.cType|isPointer %}&{%endif%}this->raw->{{ field.name }},
```
johnhaley81: Yikes...

johnhaley81:
:+1: to type safety. Thanks @srajko

PR #962, opened 2016-03-23
[WIP] Cache repository / owner values in persistent handle

srajko:
This works towards #955.  Once we free repositories on wrapper GC, any object that can return a repository (for example, a `GitCommit`) can potentially return a freed repository.  This PR adds persistent handles to any object that returns a repository.  Once we enforce a 1-1 relationship between `git_repository` objects and `GitRepository` wrappers, this will ensure that the wrapper does not get GCd until all objects that can return the repository can get GCed as well.

The stand-alone benefit of this PR is just that we don't create a new wrapper each time we return a repository from an object.

Also, the caching mechanism might be helpful beyond holding a handle - we can extend it to cache any similar value and avoid calling into the libgit2 layer for sync functions (this can help with thread safety without requiring locking).

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -1767,6 +1767,7 @@
       "ignore": true
     },
     "repository": {
+      "cacheResult": true,
       "functions": {
         "git_repository__cleanup": {
           "ignore": true
--- a/generate/scripts/helpers.js
+++ b/generate/scripts/helpers.js
@@ -179,6 +179,7 @@ var Helpers = {
     typeDef.filename = typeDef.typeName;
     typeDef.isLibgitType = true;
     typeDef.dependencies = [];
+    typeDef.cacheResult = Boolean(typeDefOverrides.cacheResult);
     typeDef.selfFreeing = Boolean(typeDefOverrides.selfFreeing);
 
     if (typeDefOverrides.freeFunctionName) {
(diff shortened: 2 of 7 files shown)


johnhaley81:
@srajko seems like it's dying on https://travis-ci.org/nodegit/nodegit/jobs/118155322#L6162-L6163

johnhaley81:
Same on appveyor

srajko:
Looks like it's creating a wrapper around a NULL raw value (a null tree in this case), and crashing trying to get NULL's owner.  I can put a check around it.

srajko:
That took care of the hangs.  There is consistent failure now with one of the submodule tests, but that's because it's using the defective repo.  Switching to a different one should help.

srajko:
Actually, no - it looks like it might be a problem with taking the result of `reference.peel`, which returns a `git_object` and then feeding it into `createBranch` which then does a `getCommit`.  It seems like it's interpreting the `git_object` as a `git_oid` inside `git_commit_lookup`.  I'm not sure why this behavior is different than before.  I'll  take a closer look at it in the morning.

srajko:
I fixed the test to pass the ID since NodeGit is expecting an ID there.  Briefly looked at the behavior on `master`, it does the exact same type misinterpretation, but for some reason it works there.

implausible:
I've completed this in master.

Issue #963, opened 2016-03-23
Bug (or documentation error) in the arguments to Diff.blobToBuffer


tylerchurch:
If you look at the argument list here for `Diff.blobToBuffer`: http://www.nodegit.org/api/diff/

And then compare that to the argument list that's actually exposed in the code; https://github.com/nodegit/nodegit/blob/83f6affb4f1305826336fd34b3126d8b35debb2a/lib/diff.js#L61

They're different, with the actual API lacking the arguments `buffer_len` (and actually computing this value itself) and `payload`.

I'd submit a fix the documentation or the API, but I wasn't sure which one was the buggy one - seems to me like computing buffer_len inside `blobToBuffer` is intentional, but it also seems odd to me (e.g. what happens if the buffer passed in isn't UTF-8? The current code assumes that it is, and I imagine it would compute an incorrect length for non UTF-8 buffers.)

johnhaley81:
So the documentation in this case is wrong. This can be fixed by doing the same that we do [here](https://github.com/nodegit/nodegit/blob/master/lib/blob.js#L7-L14) and just override the generated docs. After that's done we can close this out.

PR #964, opened 2016-03-23
diff.blobToBuffer: Fix passing in Buffers instead of strings.

tylerchurch:
Been seeing some strageness with `blobToBuffer` (see the issue I created over here, for example: https://github.com/nodegit/nodegit/issues/963 )

This pull requests fixes the fact that if you pass in an object of type Buffer, it is turned into a string, but then that string is never used (and because of the way the C bindings are written, because the object isn't of type string, it gets turned into a null, and you get a diff that looks like there's no content in the file.)

### DIFF
--- a/lib/diff.js
+++ b/lib/diff.js
@@ -84,7 +84,7 @@ Diff.blobToBuffer= function(
     this,
     old_blob,
     old_as_path,
-    buffer,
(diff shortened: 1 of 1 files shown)


joshaber:
Bah, good catch 👍

Issue #965, opened 2016-03-24
Diff.blobToBuffer crashes the node process if given a file that contains null characters


tylerchurch:
I was using nodegit to grab some diffs out of a repo, and when it hit a Visio file that was in there it would crash and die.

The Visio file was big enough and contained null characters in the right places such that it exposed a bug in nodegit that was causing libgit2 to throw a memory access violation error.

Here's some code that creates a new repo, writes out a file, tries to do a diff, and reproduces the issue:

```js
var fs = require('fs');
var path = require('path');

var rimraf = require('rimraf');
var git = require('nodegit');

var repoPath = 'C:\\blobToDiff-test-repo';
var testFilePath = "test.txt";
var testFileFullPath = path.join(repoPath, testFilePath);

rimraf(repoPath, function(err) {
    if (err) {
        throw err;
    }

    console.log("rimraf " + repoPath + " successful");

    fs.mkdir(repoPath, function(err) {
        if (err) {
            throw err;
        }

        console.log("mkdir " + repoPath + " successful");

        var repository;
        git.Repository.init(repoPath, 0)
            .then(
                function(repository_) {
                    repository = repository_;
                    console.log("repo successfully inited");
                    return repository.index();
                }
            )
            .then(
                function(index) {
                    return new Promise(
                        function(resolve, reject) {
                            var testFileContent = "Initial file content.";

                            fs.writeFile(testFileFullPath, testFileContent, function(err) {
                                if (err) {
                                    reject(err);
                                    return;
                                }

                                index.addByPath(testFilePath);
                                index.write();
                                resolve(index.writeTree());
                            });
                        }
                    );
                }
            )
            .then(
                function() {
                    console.log("file staged successfully");
                    var signature = git.Signature.default(repository);
                    return repository.createCommitOnHead([], signature, signature, "Initial commit.");
                }
            )
            .then(
                function(oid) {
                    console.log("commit successfully made: " + oid.toString());
                    return repository.getCommit(oid);
                }
            )
            .then(
                function(commit) {
                    console.log("commit object successfully gotten");
                    return commit.getTree();
                }
            )
            .then(
                function(tree) {
                    console.log("commit tree successfully gotten");
                    return tree.entryByPath(testFilePath);
                }
            )
            .then(
                function(treeEntry) {
                    console.log("tree entry successfully gotten");
                    return repository.getBlob(treeEntry.sha());
                }
            )
            .then(
                function(blob) {
[...]

PR #966, opened 2016-03-24
Switch from strdup to memcpy to faithfully copy all bytes and prevent a crash in some cases.

tylerchurch:
This is a pull request to fix the issue I created over here: https://github.com/nodegit/nodegit/issues/965

To summarize over here why I'm making these changes:

- `strdup` expects a C-style null-terminated string, but sometimes we're using nodejs strings to pass in strings that contain nulls in the middle, thus failing to copy everything at best, and causing crashes from reading invalid memory locations at worst.
- Thus I changed to using `malloc`, `memcpy` and `memset` to prevent these issues.

To explain my changes a little bit:

- I make the malloc'ed string one byte longer so I can add a terminating null character, since it looks like these templates are used in a number of different places, and for things like a repo path, where we _do_ want things to end in a null character.
- I used `memset` instead of the more standard array-style access because most of the usages of these templates are `const char *`.
- The funky casts are because some places have `void *` instead of `const char *`, so to get the pointer arithmetic right, I had to specify the `char *`.

C code isn't what I normally write, so let me apologize in advance if these changes aren't quite what they should be :smile:

I'm happy to make any modifications necessary.

The limited testing I've done seems to show that everything is working perfectly well with these changes.

### DIFF
--- a/generate/templates/partials/convert_from_v8.cc
+++ b/generate/templates/partials/convert_from_v8.cc
@@ -14,7 +14,9 @@
   {%if cppClassName == 'String'%}
 
   String::Utf8Value {{ name }}(info[{{ jsArg }}]->ToString());
-  from_{{ name }} = ({{ cType }}) strdup(*{{ name }});
+  from_{{ name }} = ({{ cType }}) malloc({{ name }}.length() + 1);
+  memcpy((void *)from_{{ name }}, *{{ name }}, {{ name }}.length());
+  memset((void *)(((char *)from_{{ name }}) + {{ name }}.length()), 0, 1);
   {%elsif cppClassName == 'GitStrarray' %}
 
   from_{{ name }} = StrArrayConverter::Convert(info[{{ jsArg }}]);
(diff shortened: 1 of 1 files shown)


tylerchurch:
Okay, not sure what I did wrong, this is the only error I see:

```
> nodegit@0.11.9 coveralls /home/travis/build/nodegit/nodegit
> cat ./test/coverage/merged.lcov | coveralls


/home/travis/build/nodegit/nodegit/node_modules/coveralls/bin/coveralls.js:18
        throw err;
              ^
Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```

Did I do that? Or is there some other error I'm missing? Or is there just some intermittent CI failure?

johnhaley81:
@tylerchurch it's an intermittent CI failure. I restarted the test. I was seeing the same on #958

johnhaley81:
Ok, that took me a moment to get my head around. Would you mind doing a summary of the comments above in the code? I know that later I'm going to look at that and go "WTF??"

johnhaley81:
Thanks @tylerchurch!

PR #968, opened 2016-03-24
Add support for thread safety on async actions only

srajko:
Adds a middle-ground thread-safety level, which locks async functions but skips locking of sync functions, to avoid the risk of blocking the main thread.

/cc @johnhaley81 @joshaber

### DIFF
--- a/generate/templates/manual/include/lock_master.h
+++ b/generate/templates/manual/include/lock_master.h
@@ -4,8 +4,15 @@
 class LockMasterImpl;
 
 class LockMaster {
+public:
+  enum Status {
+    Disabled = 0,
+    EnabledForAsyncOnly,
+    Enabled
+  };
 
-  static bool enabled;
+private:
+  static Status status;
 
   LockMasterImpl *impl;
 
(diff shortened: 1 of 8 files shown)


srajko:
It's backwards compatible with `NodeGit.enableThreadSafety()` but looks like I took away `NodeGit.disableThreadSafety()` (you would have to use the more general `NodeGit.setThreadSafetyStatus`).

srajko:
BTW the async-only thread-safety is activated using `NodeGit.setThreadSafetyStatus(1)`.  `NodeGit.enableThreadSafety()` still does the full async+sync locking (equivalent to `NodeGit.setThreadSafetyStatus(2)`)

joshaber:
Thanks @srajko!

johnhaley81:
> It might be nice to add more user-friendly identifiers instead of 1 and 2 (and 0 - disabled)

Let's make those strings that live on `NodeGit.THREAD_SAFETY` to be a little more clear.

srajko:
> Let's make those strings that live on NodeGit.THREAD_SAFETY to be a little more clear.

:+1: added.

johnhaley81:
@srajko thanks!

Issue #874, opened 2016-01-21
Add babel to let the library use ES6


johnhaley81:
Because it's 2016.

ChristianMurphy:
Awesome!
It could be worth considering moving for [jshint](https://github.com/jshint/jshint) to [eslint](https://github.com/eslint/eslint) + [eslint-babel](https://github.com/babel/babel-eslint).
JSHint currently only supports some of the syntax made available by babel.
ESLint (with the babel plugin) supports all valid babel code.

Reference: https://medium.com/@dan_abramov/lint-like-it-s-2015-6987d44c5b48#.6wprjkp4k

maxkorp:
Yeah we have a few projects already wired up well on babel+eslint

ELLIOTTCABLE:
:+1:, if only because I'm currently happily consuming NodeGit using `await/async` via Babel. It's pretty ideal.

johnhaley81:
Implemented in #1016

PR #971, opened 2016-03-26
Make index methods async

johnhaley81:
Anytime the index reads/writes to the disk it should be non-blocking. Additionally this will give us thread safety protection since async methods are piped through the `LockManager`

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -982,6 +982,12 @@
             "isErrorCode": true
           }
         },
+        "git_index_add_bypath": {
+          "isAsync": true,
+          "return": {
+            "isErrorCode": true
+          }
+        },
         "git_index_add_frombuffer": {
           "ignore": true
         },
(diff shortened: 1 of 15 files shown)


johnhaley81:
Also updated the examples to both use the async methods and remove the double reading of the index. `repo.openIndex` forces a read of the index so it doesn't make sense to do another one right after the first one happens. I think that was a holdover from an ancient version of libgit2.

joshaber:
💖

johnhaley81:
This should be gtg now. @tbranyen, @srajko, and @joshaber if you wouldn't mind :eyes: on this that would be most appreciated :)

tbranyen:
We need to start using async/await internally. ^_^

johnhaley81:
#874 :smile:

generate/input/descriptor.json
```diff
@@ -992,6 +1021,13 @@
               "isReturn": true
             }
           },
+          "isAsync": true,
+          "return": {
+            "isErrorCode": true
+          }
+        },
+        "git_index_conflict_remove": {
```
srajko: If you want to be pedantic about keeping things sorted alphabetically, this one belongs a few entries down

Issue #969, opened 2016-03-25
Nodegit doesnt work on LinuxMint (Ubuntu, I guess)


martinheidegger:
Getting this error when **running** nodegit on LinuxMint (equal to Ubuntu i guess)

```
Error: /usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.20' not found (required by /usr/lib/node_modules/nodeschool-admin/node_modules/nodegit/build/Release/nodegit.node)
```

tbranyen:
Please read the README this is discussed under installing.

martinheidegger:
Thank you for explaining, opened a pull request so future users have less problems.

PR #972, opened 2016-03-28
Error message on libstdc missing.

martinheidegger:
`nodegit` is usually used as library for other projects. Nodegit is also, unfortunately, depends on a version of libstdc++ that is not installed on many linux systems. Any person that uses nodegit as a library would need to explain this problem in their issues. In order to reduce the work of libraries/tools that depend on nodegit this PR adds an error message on postinstall if the dependency is missing and adds steps on how to mitigate that issue.

_Note: This has been a follow-up on #969._

### DIFF
--- a/postinstall.sh
+++ b/postinstall.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 if [ -n "$(node lib/nodegit.js 2>&1 | grep libstdc++)" ]; then
-	echo "[ERROR] Seems like you the latest libstdc++ is missing on your system!"
+	echo "[ERROR] Seems like the latest libstdc++ is missing on your system!"
 	echo ""
 	echo "On Ubuntu you can install it using:"
 	echo ""
--- a/postinstall.sh
+++ b/postinstall.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+if [ -n "$(node lib/nodegit.js 2>&1 | grep libstdc++)" ]; then
+	echo "[ERROR] Seems like you the latest libstdc++ is missing on your system!"
+	echo ""
+	echo "On Ubuntu you can install it using:"
+	echo ""
+	echo "$ sudo add-apt-repository ppa:ubuntu-toolchain-r/test"
+	echo "$ sudo apt-get update"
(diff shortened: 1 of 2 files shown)


tbranyen:
Love it. Means I'll have less repetitive issues to close, and indirect users will have a better idea on how to solve any issues with installing.

johnhaley81:
Testing on windows right now :)

johnhaley81:
Does windows not run the post-install script?

johnhaley81:
Ahh I didn't see the [second commit](https://github.com/nodegit/nodegit/pull/972/commits/601cacb67a19a49d5266b3122684e7d62ae6d94b) I'll confirm locally then gtg IMO.

postinstall.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+if [ -n "$(node lib/nodegit.js 2>&1 | grep libstdc++)" ]; then
+	echo "[ERROR] Seems like you the latest libstdc++ is missing on your system!"
```
johnhaley81: A little grammatical error. Should read `"[ERROR] Seems like the latest libstdc++ is missing on your system!"

postinstall.sh
```diff
@@ -0,0 +1,12 @@
...
+	echo ""
+	echo "On Ubuntu you can install it using:"
+	echo ""
+	echo "$ sudo add-apt-repository ppa:ubuntu-toolchain-r/test"
+	echo "$ sudo apt-get update"
+	echo "$ sudo apt-get install libstdc++-4.9-dev"
+	exit 1
+fi
```
johnhaley81: New line at EOF please :)

johnhaley81:
Confirmed that it works locally. I did make 2 minor comments though.

martinheidegger:
@johnhaley81 Pushed the adjustments

johnhaley81:
Thanks @martinheidegger!

paulcbetts:
This is breaking builds in Windows, you can't invoke Bash scripts in npm packages. Can you rewrite this as a JS file? Should be super easy and you can bail out if `process.platform !== 'linux'`

PR #975, opened 2016-03-28
Add support for git_repository__cleanup

srajko:


srajko:
The cleanup seems to be working nicely in conjunction with #958 - tested this in GitKraken and seeing memory usage improvements (as well as a lot more calls to `git_commit__free`) when changing repositories (and cleaning up the old repository when changing).

The advantage of `cleanup` over `free` is that `cleanup` leaves the repository object functioning, and that avoids problems with instances that still have a reference to the repo - this makes safer to call explicitly for the time being, when trying to address memory usage problems (until we get automatic memory management nailed down for all types).

johnhaley81:
LGTM, thanks @srajko

PR #977, opened 2016-03-29
Reference leak fix

srajko:
Frees references on wrapper GC, and duplicates returned oids so they are not freed with the reference.

iterator / foreach handling would need additional support, but we currently ignore all those methods.


johnhaley81:
Thanks @srajko, keep them coming :)

PR #981, opened 2016-04-01
Remote leak fix

srajko:
Makes remotes self-freeing.

I needed to add new functionality for https://libgit2.github.com/libgit2/#HEAD/group/remote/git_remote_get_refspec, because the returned refspec belongs to the remote (and gets freed with it), but there is no way to duplicate the refspec (it is an opaque struct).  AFAICT this requires an ownership mechanism, allowing the refspec to keep a handle on its owner remote so it doesn't get freed.

So I extended `shouldDuplicate` into `ownedByThis`: when `ownedByThis` is turned on we duplicate if the type has a `dupFunction`, otherwise we keep a handle on the owner.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -143,7 +143,7 @@
         },
         "git_blob_id": {
           "return": {
-            "shouldDuplicate": true
+            "ownedByThis": true
           }
         },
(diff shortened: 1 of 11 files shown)


johnhaley81:
Thanks @srajko!

PR #983, opened 2016-04-06
Factor out base class for structs / classes with cType

srajko:
Starts towards more code sharing between the struct_* and class_* combyne templates - I needed this to make the new memory management code available to structs (it was only implemented for classes initially).  There is probably more code that could be pulled in down the line but I wanted to keep this minimal for what I needed.

This leans more heavily on C++ templates than NodeGit code typically does, so this is something to consider - if needed I can probably get similar code sharing via combyne partial templates but I'd prefer to use language features when reasonable.  v8 and Nan do use C++ templates / traits classes in similar ways, so at least it's not a completely new pattern to the project.

See code / commit comments for more details on the changes.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -1307,7 +1307,7 @@
       "ignore": true
     },
     "oid": {
-      "dupFunction": "git_oid_cpy",
+      "cpyFunction": "git_oid_cpy",
       "freeFunctionName": "free",
       "shouldAlloc": true,
       "functions": {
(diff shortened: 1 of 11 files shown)


johnhaley81:
:+1:

tbranyen:
I much prefer keeping Combyne usage to a minimal, :+1: for language features.

johnhaley81:
A lot of it is navigating the treacherous world of "What does `gcc`/`MSBUILD` support in C++99/C++11".

PR #997, opened 2016-04-18
Index.conflictAdd should have nullable entries

implausible:
when you run into a conflict, sometimes one of the entries just plain old doesn't exist. An example being you create a file in 2 branches and try to merge them. They have no ancestor_entry because they were both created in their respective branches. Therefore, ancestor_entry should be null. The same can apply to our_entry and their_entry

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -998,7 +998,20 @@
           }
         },
         "git_index_conflict_add": {
-          "isAsync": true,
+          "args": {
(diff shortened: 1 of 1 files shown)


johnhaley81:
@implausible could you write a more descriptive summary of this commit please? `these are all optional parameters` doesn't give you any meaningful data. Maybe "Correct `git_index_conflict_add` optional parameters" or something?

PR #998, opened 2016-04-18
Fix a typo in function call

adius:


adius:
@johnhaley81 This is really urgent, as it completely breaks the whole entry retrieving functionality!

johnhaley81:
Confirmed, this was broken. Thanks @adius!! Could you also write some tests for this as well please?

johnhaley81:
I don't know what's going on with that last test just not starting but 8/9 greenlit so /shrug

PR #1004, opened 2016-04-25
Descriptor fixes

srajko:
Just a couple of fixes to descriptors that I wanted to submit separately.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -71,11 +71,8 @@
   },
   "types":
   {
-    "annotated": {
+    "annotated_commit": {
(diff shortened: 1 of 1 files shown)


generate/input/descriptor.json
```diff
@@ -71,11 +71,8 @@
   },
   "types":
   {
-    "annotated": {
+    "annotated_commit": {
```
johnhaley81: Can you remove this section completely? No use in having an empty section for this.

srajko: Yes, but I fill it right back up in #1006.

PR #1005, opened 2016-04-25
Ignoreinit leak fix

srajko:
Fixes a memory leak when ignoreinit was used (as described in deleted comment).


johnhaley81:
Thanks @srajko!

PR #1006, opened 2016-04-25
Oid leak fix

srajko:


johnhaley81:
Our CI has a massive backlog so I've been testing these locally. Thanks @srajko

Issue #1002, opened 2016-04-23
Diff.merge: inconsistent behavior


smith-kyle:
I have two diffs, one from `commit.getDiffWithOptions` and one from `NodeGit.Diff.treeToWorkdirWithIndex` where the tree is from the head commit. They were made with the same options. When merging the two diffs, depending on which diff is merged into which, either the diffs will merge or this error will be thrown:
> Error: Attempt to merge diffs created with conflicting options

Here's some kind of pseudo code
```javascript
let workDirWithIndexDiff;
repo.getHeadCommit()
  .then(headCommit => headCommit.getTree())
  .then(headTree => NodeGit.Diff.treeToWorkdirWithIndex(repo, headTree, {}))
  .then(_workDirWithIndexDiff => {
    workDirWithIndexDiff = _workDirWithIndexDiff;
    return someCommit.getDiffWithOptions({});
  })
  .then(someDiff => {
    // This would reject with Error: Attempt to merge diffs created with conflicting options
    someDiff.merge(workDirWithIndexDiff);

    // This would resolve
    workDirWithIndexDiff.merge(someDiff);
  });
```

smith-kyle:
When doing a diff with `git_diff_tree_to_workdir_with_index ` (which is a diff merge of `git_diff_tree_to_index` and `git_diff_index_to_workdir`) the `opts.flags` respect whatever the repo's `ignorecase` setting is. So if ignorecase is set to true for the repo, the returned diff will ignore the case.

When doing a tree to tree comparison, `ignorecase` is set to false by default regardless of the repo settings.
https://github.com/libgit2/libgit2/blob/master/src/diff.c#L1324-L1327

So AFAIK the only way to get the options to align between these two diffs when the repo's `ignorecase` is set to true would be to include the `GIT_DIFF_IGNORE_CASE` flag in the options passed into `commit#getDiffWithOptions`

smith-kyle:
I don't understand why libgit2 doesn't default to the same ignore options when calling `git_diff_tree_to_workdir_with_index` and `git_diff_index_to_workdir`

PR #1007, opened 2016-04-25
Added more Diff.merge tests

smith-kyle:
Kind of related to #1002. I'd like to eventually add a test where the diff options flags are not set to 1024 (IGNORE_CASE) but as of right now that test would fail. This is more to prove to myself that it is possible to merge a diff from a commit into a diff from the workdir and index and vice versa.

### DIFF
--- a/test/tests/diff.js
+++ b/test/tests/diff.js
@@ -349,7 +349,7 @@ describe("Diff", function() {
   });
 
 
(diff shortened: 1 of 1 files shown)


johnhaley81:
Always love more tests. Thanks @smith-kyle

PR #1008, opened 2016-04-26
Use diff options even if the commit has no parent

smith-kyle:
`Commit#getDiffWithOptions` did not use the diff options passed in if the commit did not have a parent.


smith-kyle:
This is related to #1002 in that the same error is thrown when trying to merge two diffs but this only fixes the case where one of the diffs comes from a commit without a parent

johnhaley81:
Thanks @smith-kyle

PR #1010, opened 2016-04-26
Update libgit2 v0.24.1

johnhaley81:
This updates NodeGit to use the latest stable version of libgit2. Many things are broken/changed and this will require a minor bump in the version number. Which is ok since we had to do that anyways for the next release.

This is almost an entire year of libgit2 changes so _a lot_ of stuff is fixed/changed/removed and I believe I'm even more bald now than I was when I started this thing. Which is really impressive because there's not much left.

Summary of changes that were brought in:

v0.24.0 and v0.24.1
-------

### Changes or improvements

* Custom merge drivers can now be registered, which allows callers to
  configure callbacks to honor `merge=driver` configuration in
  `.gitattributes`.

* Custom filters can now be registered with wildcard attributes, for
  example `filter=*`.  Consumers should examine the attributes parameter
  of the `check` function for details.

* Symlinks are now followed when locking a file, which can be
  necessary when multiple worktrees share a base repository.

* You can now set your own user-agent to be sent for HTTP requests by
  using the `Libgit2.OPT.SET_USER_AGENT` with `Libgit2.opts()`.

* You can set custom HTTP header fields to be sent along with requests
  by passing them in the fetch and push options.

* Tree objects are now assumed to be sorted. If a tree is not
  correctly formed, it will give bad results. This is the git approach
  and cuts a significant amount of time when reading the trees.

* Filter registration is now protected against concurrent
  registration.

* Filenames which are not valid on Windows in an index no longer cause
  to fail to parse it on that OS.

* Rebases can now be performed purely in-memory, without touching the
  repository's workdir.

* When adding objects to the index, or when creating new tree or commit
  objects, the inputs are validated to ensure that the dependent objects
  exist and are of the correct type.  This object validation can be
  disabled with the `Libgit2.OPT.ENABLE_STRICT_OBJECT_CREATION` option.

* The WinHTTP transport's handling of bad credentials now behaves like
  the others, asking for credentials again.

### API additions

* `Blob.createFromStream()` and
  `Blob.createFromStreamCommit` allow you to create a blob by
  writing into a stream. Useful when you do not know the final size or
  want to copy the contents from another stream.

* `Config#lock` has been added, which allow for
  transactional/atomic complex updates to the configuration, removing
  the opportunity for concurrent operations and not committing any
  changes until the unlock.

* `DiffOptions` added a new callback `progress_cb` to report on the
  progress of the diff as files are being compared. The documentation of
  the existing callback `notify_cb` was updated to reflect that it only
  gets called when new deltas are added to the diff.

* `FetchOptions` and `PushOptions` have gained a `custom_headers`
  field to set the extra HTTP header fields to send.

* `Commit#headerField` allows you to look up a specific header
  field in a commit.

### Breaking API changes

* `MergeOptions` now provides a `defaultDriver` that can be used
  to provide the name of a merge driver to be used to handle files changed
  during a merge.

* The `Merge.TREE_FLAG` is now `Merge.FLAG`.  Subsequently,
  `treeFlags` field of the `MergeOptions` structure is now named `flags`.

* The `Merge.FILE_FLAGS` enum is now `Merge.FILE_FLAG` for
  consistency with other enum type names.

* `Cert` descendent types now have a proper `parent` member

* It is the responsibility of the refdb backend to decide what to do
  with the reflog on ref deletion. The file-based backend must delete
  it, a database-backed one may wish to archive it.

* `Index#add` and `Index#conflictAdd` will now use the case
  as provided by the caller on case insensitive systems.  Previous
  versions would keep the case as it existed in the index.  This does
  not affect the higher-level `Index#addByPath` or
  `Index#addFromBuffer` functions.

* The `Config.LEVEL` enum has gained a higher-priority value
  `PROGRAMDATA` which represent a rough Windows equivalent
  to the system level configuration.

* `RebaseOptions` now has a `mergeOptions` field.

* The index no longer performs locking itself. This is not something
  users of the library should have been relying on as it's not part of
  the concurrency guarantees.

* `Remote#connect()` now takes a `customHeaders` argument to set
  the extra HTTP header fields to send.

* `Tree.entryFilemode`, `Tree.entryFilemodeRaw`, `Tree.entryId`, `Tree.entryName`,
  `Tree.entryToObject`, and `Tree.entryType` have all been moved to the `TreeEntry` prototype.
  Additionally, the `TreeEntry` fields have been removed in lieu of the corresponding functions to return
  the data.


johnhaley81:
image

johnhaley81:
>.< windows!!!

http://www.nooooooooooooooo.com/

johnhaley81:
💥 and we're green on CI!

johnhaley81:
@nodegit/owners anybody want to weigh in on this?

srajko:
:+1:

I'd just recommend adding the changes from https://github.com/nodegit/nodegit/pull/1010/commits/295f50703e6d995ff02de37f96c3ab91336e56b4 to the release notes, namely that `git_tree_entry_filemode`, `git_tree_entry_filemode_raw`, `git_tree_entry_free`, `git_tree_entry_id`, `git_tree_entry_name`, `git_tree_entry_to_object`, `git_tree_entry_type` got moved from being `Tree` static functions to being `TreeEntry` instance functions, and the `TreeEntry` fields are no more.

johnhaley81:
Alright, we're going live!

PR #1011, opened 2016-04-27
Add Node 6 support

johnhaley81:
~~People upgrading from 0.10/0.12 are probably going to be going to either Node 5/6 and not Node 4. Our build matrix is getting huge and I'd like to build binaries for Node 6 now. Thus I figured we could drop support for Node 4.~~

Put support for Node v4 back in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
     - export NODE_VERSION="0.12" TARGET_ARCH="x64"
     - export NODE_VERSION="4.1" TARGET_ARCH="x64"
     - export NODE_VERSION="5.8" TARGET_ARCH="x64"
+    - export NODE_VERSION="6" TARGET_ARCH="x64"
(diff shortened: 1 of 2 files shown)


johnhaley81:
Another one to run by @nodegit/owners

jdgarcia:
I think the PR title is backwards.

srajko:
Node 4 is LTS, not sure if that factors in.

johnhaley81:
Good point @srajko. I'm putting it back in then.

johnhaley81:
So instead I think that when NodeJs removes support for [v5](https://nodejs.org/en/blog/release/v6.0.0/#what-about-node-js-v5) and [v0.12](https://nodejs.org/en/blog/release/v6.0.0/#what-about-node-js-v0-10-and-v0-12) we also remove support for them.

PR #1013, opened 2016-04-27
Move reflog entry methods to ReflogEntry

srajko:
Redoing https://github.com/nodegit/nodegit/pull/1009 so it can get merged in separately.


johnhaley81:
Those were network issues on AppVeyor. This is gtg IMO.

PR #1016, opened 2016-04-28
Add babel to let the library use ES6

cbargren:
Addresses #874


johnhaley81:
m-rdb4-1313933

you just blew this up like Francis blew up that town!

tbranyen:
Why do we need to switch to ESLint? JSHint supports ES6

johnhaley81:
@tbranyen overall I'd be in favor of switching to ESLint regardless of ES6 support in JSHint. I find the docs in ESLint better overall, the linter options are fantastic and do a great job of keeping coding styling consistent as well in a project. And it's pluggable which is great. What would be the reasons to stay on JSHint (other than time spent)?

cbargren:
What @johnhaley81 is along the reasons of why I began the switch. Although admittedly, I was under the impression that JSHint wouldn't work with ES6.

cbargren:
Huh... Not sure why that test failed. It's passing for me locally.

cbargren:
Anyways, this should be good to go, once I figure out if there's anything wrong with that test. I can modify https://github.com/nodegit/nodegit/pull/1016/commits/5a094c4c6b16d0371713cb962bfeb4461f70c3ec to omit the code updates and instead modify the `.eslintrc.js` to more closely match the old JSHint if we'd rather have it that way. Of course that all is dependent on where we want to go WRT JSHint versus ESLint.

tbranyen:
Some other concerns:

- Can any ES6 code changes wait for a separate PR. As this PR stands, part of the code base is using ES6 modules and syntax while the rest is not.

- Can we avoid using ES6 modules completely for now? Node will have it's own loader specification soon and we should wait and align with that. I don't see the benefits of mixing them with CJS.

- What are the debugging implications? I'd imagine Node doesn't consume source maps, or does it?

- As for ESLint vs JSHint, it's a moment of "not broken, don't fix". If we want style rules, we should use a tool designed for it (JSCS). Even though it's planning on merging with ESLint, I'd imagine a standalone will always exist. I've had terrible experiences with JSCS upgrading, so I would recommend just locking to the latest stable they've released.

tbranyen:
Another thought is to only support ES6 features that natively work and provide the Babel transpile as a fallback purely for Node 0.12 or whatever minimum version we set.

cbargren:
@tbranyen 
- Yeah, I can definitely remove all code changes from this PR. 
- I don't see any problem with avoiding using ES6 modules yet. I hadn't planned on switching any of that over and I think the primary reason for adding ES6 support was some of the other neat stuff that it does. 
- node-debug does work with source maps and babel can also generate them
- ESLint is designed for both linting and style rules. Pulling in another tool to handle something that can be just as easily handled with one tool seems counterproductive, IMO.

maxkorp:
@tbranyen So just don't use anything requiring a runtime shim?

I mean, to be honest I'm not sure i follow the pushback. Is there any particular reason to avoid use of newer es6 features (not including module loading)? I guess runtime shims, but fortunately not too much requires those anymore. Totally agree wrt not changing anything in this same PR though.

Re "If we want style rules, we should use a tool designed for it (JSCS)". Thats pretty much exactly what we use jshint for. While i prefer eslint, I'm fine not switching. That said, i wouldn't bother switching to jscs since theyve merged with eslint already. When you talk about a "standalone", what does eslint have that jscs doesnt? This feels like a very odd, probably academic distinction, unless I'm missing something there.

tbranyen:
@cbargren I think you forget that JSHint at one time also did some style checks, but once JSCS became better suited for it, was able to drop those rules and focus entirely on linting (bug avoiding).

maxkorp:
Is there a tangible benefit to seperate bug-avoidance linting and style linting two a tool that does one, and a tool that does both to handle the second part, instead of just using the tool that does both to do both?

tbranyen:
@maxkorp I'm totally interested in discussing ESLint, but I'm starting to see that it's a separate discussion not related to ES6. Unless ESLint can do something JSHint can't when it comes to introducing ES6, I think it should probably be a separate PR.

maxkorp:
Yeah, inclined to agree 👍

cbargren:
Cool, I'll create a separate PR for ESLint and roll this one back before ESLint was added. 👍

tbranyen:
Okay I think besides that, I'm cool with the change. My reasoning for only using syntax supported in the runtime is that it'll be so much easier to drop Babel later on. I'd suggest not using any non-standard features even tho Babel transpiles them just fine. (Like module loading).

tbranyen:
@cbargren if you find during that rollback that JSHint simply doesn't work for some reason with `esnext: true` set, that would be a good reason to keep eslint in.

cbargren:
@tbranyen Sounds good, thanks. BTW, https://github.com/nodegit/nodegit/pull/1016/commits/d31e7c142822b6a867069ac6cfeec9a31229de31 addresses some code that is technically broken but still works before ES6, but will not work when we start using ES6. Do you want me to PR this commit separately?

tbranyen:
@cbargren I think that shows a good reason for adding ES6, I'd keep it in :+1:

cbargren:
Looks like AppVeyor died, I'm pretty sure that failure isn't valid. :(

johnhaley81:
I'm not sure why it's dying though. It's trying to rename a `package.json`?

cbargren:
It's an old bug in npm, if I'm not mistaken: https://github.com/npm/npm/issues/9696

cbargren:
It's fixed in NPM 3.6: https://github.com/npm/npm/releases/tag/v3.6.0

cbargren:
After rerunning the tests ~6 times with virtually no actual changes, they are now passing. AppVeyor/Travis was heeere. Lemme know if there's anything else needed for this. 👍

lifecycleScripts/install.js
```diff
@@ -63,7 +64,45 @@ function prepareAndBuild() {
...
+  };
+  var home = process.platform == "win32" ?
+              process.env.USERPROFILE : process.env.HOME;
+
+  opts.env.HOME = path.join(home, ".nodegit-gyp");
+
+  return new Promise(function(resolve, reject) {
+    var child = cp.spawn(cmd, args, opts);
+    child.on("close", function(code) {
+      console.log(code);
```
tbranyen: Is this line intentional?

tbranyen:
One super tiny thing and then I'm all :sparkling_heart:

lifecycleScripts/install.js
cbargren: Whoops, nope, that's just a bit of copypasta. Removing.

johnhaley81:
A few scattered failures due to networking issues on CI. LGTM. Thanks @cbargren

tbranyen:
Awww yea! Let the refactoring begin :D

PR #1017, opened 2016-04-29
Make libgit2 an actual submodule

johnhaley81:
This PR should hopefully help us track how libgit2 changes from version to version by making it an actual submodule of the library.

I added scripts to automate updating any submodules of the repo so hopefully if we add more submodules (which I think I will) they should just be dropped in and work (famous last words).

Currently we do the following for updating submodules during a `lifecycleScripts/install`:

 - If you have a submodule that is not initialized we will initialize it recursively for you.
 - If you have a submodule that is initialized we don't do anything to it.
 - If you have a submodule that is on a different commit we warn you about it in the console and don't modify the submodule.
 - If you have uncommited changes to a submodule we stop building/compilation and notifiy the user that they need to resolve this before we continue.

During a `generate` we just check to see if the submodules are all good (i.e. all initialized, on the correct commis and no work dir changes) and if any aren't we'll put up a warning in the console.


johnhaley81:
@tbranyen @maxkorp thoughts?

maxkorp:
Cant rely on git being on the path, especially on windows :(

johnhaley81:
@maxkorp yeah we prompt in that case.

PR #1023, opened 2016-05-03
[WIP] Use ESLint with a config base of eslint-airbnb-config-base

cbargren:
This replaces JSHint with the newer, more powerful ESLint. Some discussion has been had about this already at https://github.com/nodegit/nodegit/pull/1016#issuecomment-215585622.


cbargren:
So I think it'll be a good idea to start `babel`ing the test/utils/lifecycleScripts folders as well while I'm in here. This is inspired by the fact that https://github.com/nodegit/nodegit/pull/1023/commits/db56d3686e127a1b9d604ec968007cb56ebfa361 fixes the linter but is actually not fully supported as the `0oXXX` octal syntax is new to ES6. Rather than use some weird parseInt magic, We can just go ahead with babeling the test folder so we don't have to worry about what syntax is supported yet. My thoughts for the new babel structure (with the help of @implausible) are as follows:

- All un-transpiled javascript will move into a `js` folder at the root which will now contain lib, lifecycleScripts, test and utils. The generate folder could also be a candidate, though that will probably be a bit more complicated.
- The `js` folder will be transpiled to the parent directory, putting the output folders exactly where they used to be. These folders will now be `.gitignore`ed and the new folders will be `.npmignore`ed.

The biggest problem I see here is that any outstanding js changes that will go in immediately after this are going to be a merge conflict nightmare. We'll have to work this in quickly after we agree on something to minimize immediate conflicts.

Thoughts?

johnhaley81:
Why `js/lib/` -> `lib/` and not `lib/` -> `dist/` like we do now?

js/test/tests/diff.js
```diff
@@ -245,12 +245,12 @@ describe("Diff", function() {
 
   it("can diff the contents of a file to a string with unicode characters",
     function(done) {
+    this.timeout(0);
```
johnhaley81: Is this needed?

johnhaley81:
Oh, ok I'm following you guys now. That looks good to me.

johnhaley81:
@tbranyen what do you think?

js/test/tests/diff.js
cbargren: Oops, I thought I took that out. I'll take that out.

tbranyen:
Changing the folder structure will be a rebase nightmare as previously outlined.  I don't see a huge rush to bring this in immediately, but I'll try and review this weekend! Thanks for humoring me with this exercise, I know it's not the fastest thing to land, but it affects a lot.

johnhaley81:
This is definitely *not* time critical :)

cbargren:
Awesome! Yeah, the rebase nightmare for any pending PRs is definitely the biggest concern here. Thanks for checking it out!

johnhaley81:
Well, this is conflicting with like 100 files so @cbargren are you able to rebase this and clean it up?

maxkorp:
At this point, the actual file changes probably need to be reworked. FYI passing `--fix` to eslint helps with a LOOOOOT of the trivial stuff

cbargren:
I'll have to take a look at it again when I have some time. Yeah, a lot of these changes will probably have to be completely reworked, but it's really only worth doing if we're committed to getting it in immediately when it's ready, since the longer it waits, the more rebase nightmares it's going to create.

gucong3000:
Should we just lint files that changed?

gucong3000:
https://www.npmjs.com/package/polyjuice

PR #1030, opened 2016-05-11
Memory management: plugged leak in fastWalk

smith-kyle:
Thanks @srajko for holding my hand through this fix 👬


johnhaley81:
LGTM. Thanks @smith-kyle

PR #1033, opened 2016-05-12
Set oid.selfFreeing to true instead of "true"

smith-kyle:
Does not modify code generation since previously the string "true" evaluated to truthy


tbranyen:
Ah okay good catch. This could have caused problems if we were using `===`

johnhaley81:
Thanks @smith-kyle

PR #1036, opened 2016-05-17
Memory management: Self freeing fix

smith-kyle:
If trait is owned and is non-duplicable, disallow it from marking itself as selfFreeing


johnhaley81:
Thanks again :)

PR #1047, opened 2016-06-08
Replace then() with done(), so that error callback is not swalled by …

wasadigi:
When using the revwalk.walk function, if I make an error in the implementation of my callback, it is swalled by the code. Replacing then() with done() fixes the issue.


johnhaley81:
This function needs a better implementation as well. Someday :)

Thanks @wasadigi

PR #1055, opened 2016-06-14
Bring life cycle in line with npm standards

johnhaley81:
We were doing our own highly custom (read: fragile) life cycle scripting for installing NodeGit.

This should make it more in line with npm standards.


johnhaley81:
@tbranyen @maxkorp we're passing CI. Can you guys give some 👀 on this?

tbranyen:
I'll test it tonight :heart_eyes:

maxkorp:
Talked with john about this in person a moment ago, but putting here for records. 

We can't clean the source afterwards.

The standard workflow for electron/nwjs is to npm install normally, then run a rebuild with the appropriate headers. This works with all other native modules i've seen, and currently works with nodegit, but cleaning the src/includes out would break that. We have to be able to run `npm run rebuild` after install has finished. 

Otherwise, gonna review it up right now.

.npmignore
```diff
@@ -1,11 +1,14 @@
+/.travis/
 /build/
 /examples/
+/generate/
+/guides/
+/lib/
```
maxkorp: Why ignore lib? all of our js is in there?

generate/scripts/generateJson.js
```diff
@@ -222,7 +222,7 @@ module.exports = function generateJson() {
     _.merge(enumerable, _.omit(override, ["values"]));
 
     output.push(enumerable);
-  }).value();
+  });
```
maxkorp: don't we need the `.value` to make lodash actually do the operation? I thought they were lazy about that as of v4?

.npmignore
johnhaley81: lib should already be 100% built and in dist.

generate/scripts/generateJson.js
johnhaley81: Correct. We needed it in v3 but in v4 a `forEach` just executes.

maxkorp: PRAISE BE TO LODASH, FOR THEY ARE IMPROVE

.npmignore
maxkorp: That makes complete and total sense and I should have realized that, but sometimes I am do the forget.

lifecycleScripts/postinstall.js
```diff
@@ -20,13 +20,33 @@ function printStandardLibError() {
...
+      }
+    })
     .then(function() {
       // Is we're using NodeGit from a package manager then let's clean up after
       // ourselves when we install successfully.
       if (!buildFlags.mustBuild) {
-        fse.removeSync(path.join(rootPath, "vendor"));
-        fse.removeSync(path.join(rootPath, "src"));
-        fse.removeSync(path.join(rootPath, "include"));
+        // We can't remove the source files yet because apparently the
```
maxkorp: This is insufferably true :(

maxkorp:
Good change, better comment.

maxkorp:
This should be good now :)

package.json
```diff
@@ -51,11 +52,7 @@
     "js-beautify": "~1.5.10",
     "jshint": "~2.8.0",
     "lcov-result-merger": "~1.0.2",
-    "lodash": "~3.10.1",
-    "mocha": "~2.3.4",
-    "nan": "^2.2.0",
-    "node-gyp": "~3.0.3",
```
maxkorp: The gyp exclusions make mme SO GODDAMNED HAPPY.

tbranyen:
I'm so far removed from the lifecycle hooks code lately, but this looks v. good, nice work!

maxkorp:
@tbranyen There are a lot of improvements in npm3, but there are some definite regressions, and it doesn't fix the fact that npm2 is bundled with node 4 which is LTS :/

By excluding having to generate the code in our bundles, we just have to configure libssh2 if needed, ensure our deps are all installed when we run node-pre-gyp (this is not guaranteed properly in npm2, but it is in npm3 and trying to insure it ourself anyways actually breaks in npm3). We fall back to the default node-gyp now, since all that confusion was from A) node 0.8 and then came back in iojs 1/2. 

A lot has changed, but overall the process is easier to follow now in npm3.

tbranyen:
:tada:

Issue #1057, opened 2016-06-17
Some examples outdated


jorisroling:
I found that the examples refer to ```entry.filename()``` where they should have referered to ```entry.name()```.

These files should be changed:
* examples/read-file.js 
* examples/clone.js 
* examples/general.js

tbranyen:
Can you submit a PR?

PR #1058, opened 2016-06-21
changed references to entry.filename() to entry.name()

jorisroling:
In response to https://github.com/nodegit/nodegit/issues/1057#issuecomment-226798781


johnhaley81:
Thanks for the PR!! <3

PR #1071, opened 2016-07-15
[WIP] Bump libssh2 to 1.7.0

maxkorp:
Tests are all screwy for me locally (including on master), so we'll see if this works on CI, but it _seems_ all good.


maxkorp:
Definitely appears to have some windows issues, probably missing and/or superfluous includes.

PR #1079, opened 2016-07-19
fix Tree#entryByName function and add test

novalis:


tbranyen:
Looks good to me :+1:

PR #1080, opened 2016-07-19
add isSubmodule() method

novalis:


novalis:
Pretty sure my commit didn't break this test, as it is in an unrelated area.

maxkorp:
Nice 👍

PR #1081, opened 2016-07-19
Fix two typos in the documentation for Commit#parents:

novalis:
1. It doesn't take a callback
2. Possessive form needs a single quote instead of a double quote.


maxkorp:
TY!

PR #1090, opened 2016-07-26
Memory management: duplicate time

smith-kyle:
It's unsafe to use the signature's time directly since the signature could be freed. Instead duplicate the signature's time when it is requested and set it to `selfFreeing`.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2453,6 +2453,9 @@
         }
       }
     },
+    "time": {
+      "dupFunction": "git_time_dup"
+    },
     "trace": {
       "functions": {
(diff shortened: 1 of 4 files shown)


srajko:
This is great, thanks @smith-kyle!  Could you add a test confirming that the time object no longer gets corrupted (for example after getting a time using `Signature.when`, and then the signature getting freed) and that the duplicated time objects do get freed now?  If you search for `garbageCollect` you can find some tests that do similar things.

smith-kyle:
Thanks @srajko I just added a test. I also confirmed that this test fails on master

srajko:
LGTM.  The travis failure seems network related.

PR #1091, opened 2016-07-26
Generate script: Don't write .cc or .h files if unchanged

smith-kyle:
The generation script now checks whether the file has changed before writing it to `/src` or `/include`. This is to improve compilation time when testing changes to generated code.

It works by creating a `/temp` directory, writing the generated code to `/temp/src` and `/temp/include`, then syncing those folders with `/src` and `/include` by deleting files  that no longer exist and copying files that have changed or been added since the last code generation. Finally the `/temp` directory is deleted

If `/src` and `/include` don't exist (i.e. it's the first time the generation script has been run), `/temp/src` will be copied to `/src` and `/temp/include` will be copied to `/include`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,6 @@
 /lib/nodegit.js
 /node_modules/
 /src/
-/temp/
 /test/coverage/
 /test/home/
 /test/repos/
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /lib/nodegit.js
 /node_modules/
 /src/
+/temp/
 /test/coverage/
 /test/home/
 /test/repos/
--- a/generate/scripts/generateJson.js
+++ b/generate/scripts/generateJson.js
@@ -231,7 +231,7 @@ module.exports = function generateJson() {
   }
 
 
-  utils.writeFile(path.join(__dirname, "..", "output", "idefs.json"), output);
+  utils.writeLocalFile("output/idefs.json", output);
 };
 
 if (require.main === module) {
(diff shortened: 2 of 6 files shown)


maxkorp:
You realize that using `node-gyp build` instead of `node-gyp rebuild` already takes care of this, right?

implausible:
@maxkorp i think this fixes the issue where you make a change to the templates / add definition / add supplement, and you need to run generate, again. This stops the generate script from overwriting unchanged files. That way if you run node-gyp build, it will only build the files that have changed, rather than compile everything @smith-kyle am I right here?

maxkorp:
Ah interesting, my bad. Lemme play with it a bit, that could be useful then :+1:.

maxkorp:
So, the overall recompile time without this is still fairly small, the bulk of the overall compile time is openssl and that isn't affected in this case (and the second largest is libgit2, which is also unaffected). 

That said, if this is a big deal to you guys I'm cool with it. Any reason we don't use `os.tmpdir` though?

smith-kyle:
Still relatively small but it's going from ~90 seconds to < 5 seconds if you've only changed a couple generated files.

This is the first I'm hearing about `os.tmpdir`. I don't know much about temp dir etiquette but this seems like a good use for it. Thanks for the suggestion @maxkorp I'll switch it over.

maxkorp:
Awesome. Thanks a pretty big change, so I can dig it :+1:

smith-kyle:
@maxkorp I've (finally) moved this to use `os.tempdir`. It's ready for review. Not sure why appveryor is dying

maxkorp:
I don't have push access to your repo to rebase this, so you gotta rebase this, but it's G2G man.

smith-kyle:
Rebased 👍

johnhaley81:
Tested. Looks good, works good. Thanks @smith-kyle!

PR #1107, opened 2016-08-11
Try re-enabling node 6.3

maxkorp:
6.3 is working for me locally now, so trying this here.


maxkorp:
Interestingly, this works for me locally now (I got the same break appveyor is showing before, but now i don't get that with my last change).

maxkorp:
Depends on https://github.com/nodegit/nodegit/pull/1125

Not comfortable calling this the correct long term fix (not sure if it links to the built in or our own when both are available) but better than not building at all. Thanks to @implausible for helping find this.

Issue #1045, opened 2016-06-06
Not possible to getEntry of commit returned by fileHistoryWalk.


martinheidegger:
When I tried to do the following for an existing file in my repository:

```javascript
walker = repo.createRevWalk()
walker.sorting(git.Revwalk.SORT.TIME);
walker.push(headCommit)
walker.fileHistoryWalk(path, 10000).then(function (commitSets) {
   commitSets[0].commit.getEntry(filePath)
})
```

but this throws the error `TypeError: Cannot read property 'getTree' of undefined`.

When examining the commit I saw that the commit is without a repo?!

nodegit version: `0.13.1`

johnhaley81:
That is a manual function added to `RevWalk` that doesn't create commits via our JS wrapper function so it doesn't get some stuff that's necessary on it. A work around for it right now is to grab the commit's `sha` and then re-grab the commit which will then have the info on it that you need but clearly the C++ function needs to be updated. 

@implausible can you take a look at this when you get some time?

littlebee:
I was able to work around this by setting commit.commit.repo = repo I used to create rev walker

implausible:
@johnhaley81 This just needs an additional step at the Javascript layer. In the lookupwrapper util, we add repo to the commit objects ourselves. I'll fix it this morning.

rcjsuen:
I got hit by this also.

PR #1110, opened 2016-08-12
Fix getEntry for commits retrieved by fastHistoryWalk

implausible:
Addresses: https://github.com/nodegit/nodegit/issues/1045

### DIFF
--- a/lib/revwalk.js
+++ b/lib/revwalk.js
@@ -27,7 +27,7 @@ var _sorting = Revwalk.prototype.sorting;
  */
 Revwalk.prototype.fileHistoryWalk = function(filePath, maxCount) {
   var repo = this.repo;
-  return _fileHistoryWalk(filePath, maxCount)
+  return _fileHistoryWalk.call(this, filePath, maxCount)
     .then(function(historyEntries) {
       return historyEntries.map(function(historyEntry) {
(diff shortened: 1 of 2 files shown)


lib/revwalk.js
```diff
@@ -16,14 +18,23 @@ var _sorting = Revwalk.prototype.sorting;
...
  * @param {Number} max_count
  * @async
  * @return {Array<historyEntry>}
  */
-Revwalk.prototype.fileHistoryWalk = fileHistoryWalk;
+Revwalk.prototype.fileHistoryWalk = function(filePath, maxCount) {
+  var repo = this.repo;
+  return _fileHistoryWalk.call(this, filePath, maxCount)
+    .then(function(historyEntries) {
+      return historyEntries.map(function(historyEntry) {
```
maxkorp: Why map this? it's a referential assignment, just do a foreach?

johnhaley81: Best IMO would be: `.then(fp.map(fp.set(['commit', 'repo'], repo)))`

implausible: @johnhaley81 that's a great idea :). It is done.

jaywcjlove:
@implausible Hope to merge.

PR #1113, opened 2016-08-17
Added instructions for CircleCI users

jamesbibby:
The Circle CI instructions should be pretty obvious given the Ubuntu instructions, but this might save some people forgetting the -y flags or no knowing which part of the circle.yml to do this install in.


tbranyen:
lgtm, thank you!

Issue #905, opened 2016-02-16
Update docs on how to compile for an electron app.


geowarin:
Hey guys. Thank you for your amazing work, nodegit is a fine piece of Software.

I'm asking for your advice here.
I managed to include nodegit in our electron app but since it depends on C binaries, it basically forces us to compile the application separately on each target.

That's a shame because we lose the possibility to create all the targets clients (linux, mac and windows) in one build on a single machine.

There must be some kind of way of linking pre-compiled binaries on the fly but I could really use your help here!

Many thanks in advance.

tbranyen:
You could probably script out a solution that involves curling down all our tarballs and installing via `node-pre-gyp`. We could help brainstorm some ideas, but you may want to consider taking the same route we are, which is building the prebuilts during continuous integration.

maxkorp:
When building Gitkraken we use pre-built binaries, via an npmrc. You can also just manually force the config values with environment variables you set for the install. Using the prebuilts is highly recommended, as there are lots of finicky bits for linux, especially around openssl, especially if you're distributing this externally. 

This is our .npmrc 
```
runtime = electron
target = 0.36.5
target_arch = x64
disturl = https://atom.io/download/atom-shell
```

That way, ANY modules that have to be recompiled will automatically build for electron right away, instead of for node first, then rebuilding for electron. Additionally, node-pre-gyp will know automatically not to download node binaries, but will download electron binaries if available.

maxkorp:
I really need to update the how to for electron >.<
Also, i wish more workflows for electron would just adopt this in general, because it works so friggin well, even without nodegit involved.

geowarin:
Thanks for your awesome suggestions!
@tbranyen, moving the packaging is a very relevant and simple idea.

@maxkorp's solution is more intriguing (I had a look at gitkraken, BTW, it's awesome).
Let me rephrase it to make sure I understand:

When installing nodegit, node-pre-gyph will run. It can optionally pick up flags from the command-line or in .npmrc.
So the build would become:

1. `npm i nodegit --target_platform=win32`
2. package the electron app as usual

maxkorp:
:+1:

geowarin:
Thanks a lot @maxkorp, your solution works flawlessly!

For those interested here is our [packaging script](https://gist.github.com/geowarin/93e7aaef640c6f90d9e4).

We decided to build the app with npm 2, less [headaches](https://github.com/npm/npm/issues/9674).

[...]

PR #1114, opened 2016-08-18
fix up electron and nw.js docs

maxkorp:
Should resolve https://github.com/nodegit/nodegit/issues/905, as well as comparable fixes for nw.js


maxkorp:
Only failure was a network failure, and this is a docs only change.

PR #1125, opened 2016-09-14
patch libssh2 to work with vs2015

maxkorp:
When building with the windows SDK for vs2015, the build would die due to some duplicate defines. 

Found a variation of this patch on github.com/libssh2/libssh2 for libssh2 1.7, which we still need to update to, but for now this works.

### DIFF
--- a/vendor/libssh2/win32/libssh2_config.h
+++ b/vendor/libssh2/win32/libssh2_config.h
@@ -23,6 +23,7 @@
 #define HAVE_SELECT
 
 #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


saper:
One important thing is that the C++ code should not be built with platformtoolset v140 (default vs2015 mode) until node itself switches over, when using vs2015 one should use platformtoolset v120, which is vs2013 compatibility. The vs2015 module may appear to work at first but it might start to fail in some obscure and non-obvious situations requiring runtime C++ support! like exception stack unwinding or thread local storage,

PR #1126, opened 2016-09-14
Fix CI

maxkorp:
preinstall calls generate when running locally in a git repo, which requires fs-extra, which is not present during preinstall. I have 0 idea how this was able to work before.

Moved it to install, should behave the same is it would currently with the note that it now works on CI (or any other time you fresh clone)


maxkorp:
This is getting hung up thanks to network latency on travis, but it's working otherwise, so I'm going to merge it so other PRs can get started.

PR #1135, opened 2016-09-23
Add "Path" to ssh variable names in tests for clarity

kokarn:
Because the SSH stuff is documented with the tests at the moment it helps out if the var names are a bit clearer.


johnhaley81:
Thanks @kokarn

PR #1136, opened 2016-09-23
Fix object memory leaks

srajko:
Addresses leaks of `GitObject` instances.  Added test, also tested in GitKraken for stability.


johnhaley81:
@srajko so it seems that this is causing the freeing of a `Commit` object to break now on some platforms.

PR #1137, opened 2016-09-23
Fix revwalk memory leaks

srajko:
Addresses leaks of GitRevwalk objects. Added a test and tested in GitKraken for stability.


johnhaley81:
👍

PR #1143, opened 2016-10-04
Fix windows exception handling

srajko:
Even though we had the `/EHsc` option in the `gyp` file, it was not being applied correctly.  With the change, the option does get applied, which also takes care of numerous build warnings like this one:
```
C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\xtree(839): warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc (..\src\lock_master.cc) [C:\projects\nodegit\build\nodegit.vcxproj]
```

### DIFF
--- a/generate/templates/templates/binding.gyp
+++ b/generate/templates/templates/binding.gyp
@@ -80,26 +80,11 @@
                 "AdditionalOptions": [
                   "/EHsc"
                 ]
-              }
-            },
-            "configurations": {
-              "Debug": {
-                "msvs_settings": {
-                  "VCLinkerTool": {
-                    "AdditionalOptions": [
-                      "/FORCE:MULTIPLE"
-                    ]
-                  }
-                }
               },
-              "Release": {
-                "msvs_settings": {
-                  "VCLinkerTool": {
-                    "AdditionalOptions": [
-                      "/FORCE:MULTIPLE"
-                    ]
-                  }
-                }
+              "VCLinkerTool": {
+                "AdditionalOptions": [
+                  "/FORCE:MULTIPLE"
+                ]
               }
             }
           }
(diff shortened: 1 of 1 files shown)


saper:
You can use ExceptionHandling: Sync in gyp to indicate /EHsc

srajko:
@saper thanks for the tip, that would be cleaner but it looks like it still results in warnings for node 0.12 builds.  You can compare https://ci.appveyor.com/project/TimBranyen/nodegit/build/3029 (using `"ExceptionHandling": "Sync"`) and https://ci.appveyor.com/project/TimBranyen/nodegit/build/3031 (using `"AdditionalOptions": [ "/EHsc" ]`).

Maybe when we drop node 0.12?

saper:
My mistake. The parameter should be integer, I think the value of one(1) should do it.

saper:
`/FORCE:MULTIPLE` also has its gyp equivalent, `ForceFileOutput`.

srajko:
I am getting the same results with `"ExceptionHandling": 1` as with `"ExceptionHandling": "Sync"` - the `/EHsc` warning disappear for node 4, 5, and 6, but remain for 0.12.  This is the build output for `"ExceptionHandling": 1`: https://ci.appveyor.com/project/TimBranyen/nodegit/build/3032

So both `1` and `Sync` have an effect for node >0.12 but not 0.12.  See in contrast the build output for an unrelated PR, where the `/EHsc` warnings appear for all versions of node: https://ci.appveyor.com/project/TimBranyen/nodegit/build/3030

saper:
Strange, it should not depend on the node version. I'll check when I get somewhere near Windows. In the meantime go ahead with your changes. I just figured out at node-sass we are also using /EHsc literally.

PR #1144, opened 2016-10-04
Fix CI failures on node 0.12

srajko:


srajko:
The failure with node 0.12 was happening after `npm install`, but it doesn't look like the failure was critical to the rest of the build.  The fix I converged on ignores the `npm install` failure for node 0.12 (see comments in the `yml` file).  If this is acceptable, I can squash this down to merge.

johnhaley81:
This looks good to me. We're going to stop supporting 0.12 in December anyways since that's it's EOL but this will let us get to that mark at least.

PR #1148, opened 2016-10-14
Fix postinstall script crash when install location has spaces in its path

jon-hall:
After installing `nodegit` in a directory which had spaces along its path, the postinstall script crashed as a result (`Error: Cannot find module 'C:\<path up to the first space>'`) - wrapping this `node` call's target in quotes fixes the issue.

### DIFF
--- a/lifecycleScripts/postinstall.js
+++ b/lifecycleScripts/postinstall.js
@@ -30,7 +30,7 @@ module.exports = function install() {
     return Promise.resolve();
   }
 
-  return exec("node " + path.join(rootPath, "dist/nodegit.js"))
+  return exec("node \"" + path.join(rootPath, "dist/nodegit.js\""))
     .catch(function(e) {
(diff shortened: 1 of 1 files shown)


lifecycleScripts/postinstall.js
```diff
@@ -30,7 +30,7 @@ module.exports = function install() {
     return Promise.resolve();
   }
 
-  return exec("node " + path.join(rootPath, "dist/nodegit.js"))
+  return exec("node \"" + path.join(rootPath, "dist/nodegit.js\""))
```
johnhaley81: I feel like `return exec("node \"" + path.join(rootPath, "dist/nodegit.js\""))` should instead be `return exec("node \"" + path.join(rootPath, "dist/nodegit.js") + "\"")` but this has been open for long enough. I'll fix that in a later PR.

Issue #1160, opened 2016-11-18
npm install fails on freebsd 11.0-RELEASE-p2


Kanaye:
npm install nodegit fail when trying to build libgit2.

The errors generated:
```
  CC(target) Release/obj.target/libgit2/vendor/libgit2/src/iterator.o
../vendor/libgit2/src/iterator.c:1459:44: error: no member named 'st_ctime_nsec' in 'struct stat'
        iter->entry.ctime.nanoseconds = entry->st.st_ctime_nsec;
                                        ~~~~~~~~~ ^
../vendor/libgit2/src/iterator.c:1462:44: error: no member named 'st_mtime_nsec' in 'struct stat'
        iter->entry.mtime.nanoseconds = entry->st.st_mtime_nsec;
```

System is FreeBSD 11.0-RELEASE-p2
Node version: v7.1.0
npm version: 3.10.9

The npm-debug logfile and the stderr are attached below (the compiler errors / warnings seems not to get captued by npm).

Building the submodule in ``vendor/libgit2`` as documented on the libgit2 repo works.

Please let me know if you need more information.

Thanks for your time.

[npm-debug.log.txt](https://github.com/nodegit/nodegit/files/599757/npm-debug.log.txt)
[npm-install-stdderr.log.txt](https://github.com/nodegit/nodegit/files/599758/npm-install-stdderr.log.txt)

saper:
This command to compile `src/iterator.c` fails:

````
[...]

PR #1165, opened 2016-11-26
FreeBSD also uses struct timespec st_mtim

saper:
Fixes https://github.com/nodegit/nodegit/issues/1160


jariz:
Any updates, maintainers?
Without this PR, nodegit isn't working for an entire platform, which I'd say, is kind of important?

PR #1172, opened 2016-12-05
Add some comments on instance variables

novalis:


novalis:
See also https://github.com/nodegit/nodegit.github.com/pull/45

novalis:
Ping on this and https://github.com/nodegit/nodegit.github.com/pull/45 ?

novalis:
Pinging again.

johnhaley81:
Hi @novalis sorry about the lack of response for so long. The issue with this PR is that the `libgit2-docs.json`  is nuked anytime we update libgit2 so this change will get clobbered. Instead we do right now is add things in with the `supplement.json` file so those changes are persisted between libgit2 updates. I don't know if we have any code that pulls in the comments for things but that's where I would start. Having these in the final generated docs would be 💯

novalis:
It's been so long since I looked at this that I don't even remember how to go through whole the round trip to generate the docs, but this might work.

novalis:
... but even if not, https://github.com/nodegit/nodegit.github.com/pull/45 is still probably a good idea.

johnhaley81:
Pretty much clone `nodegit.github.com` and symlink nodegit into `nodegit.github.com/generate/nodegit` then `script/bootstrap && script/server`

novalis:
Over the weekend, I re-ran this and fixed a typo.  I think it's ready.

Looks like the AppVeyor build might be flaky?  It's failing on a bunch of PRs with an internal compiler error.

Thanks.

PR #1178, opened 2017-01-03
Add @async tag to getEntry

rcjsuen:
Fix the documentation generation for Tree's getEntry method so that it'll be properly tagged as ASYNC in the [output](http://www.nodegit.org/api/tree/#getEntry).

This fixes [nodegit.github.com issue #41](https://github.com/nodegit/nodegit.github.com/issues/41).


johnhaley81:
Thanks @rcjsuen!

PR #1179, opened 2017-01-03
Fix incorrect anchor link in TESTING.md

rcjsuen:
Changed a link to point internally instead of unnecessarily using an
external link.


saper:
Thank you!

PR #1181, opened 2017-01-06
Added tests for Branch

rcjsuen:
```
move
lookup
createFromAnnotated
```

Note that the test for `createFromAnnotated` is failing for me. It seems like the branch is being created even when the `force` flag has **not** been set. I would have expected the conflict in the branch's name to prevent the tag from being created but that's not happening.

The returned annotated commit does have the correct SHA-1 hash so that is not the issue. The branch that gets returned by the promise from `createFromAnnotated` has the same SHA-1 hash as what it was originally pointing at (ergo, the master commit) so it's almost as if the function call never happened! :thinking:

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -183,6 +183,7 @@
           }
         },
         "git_branch_create_from_annotated": {
+          "isAsync": true,
           "args": {
             "ref_out": {
               "isReturn": true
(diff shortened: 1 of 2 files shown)


rcjsuen:
I realized that the error was due to the function being synchronous and always returning something regardless of whether the function succeeded or failed. This caused the `then(*)` function to be called even though it had actually failed (which then caused the test to fail). Changing the function to be asynchronous now gets the test to pass successfully.

johnhaley81:
Thanks @rcjsuen!

PR #1190, opened 2017-01-17
Checkout's tree(*) function does not support Oid as a parameter

rcjsuen:
The docs incorrectly spec that an `Oid` can be used as a `treeish` but the libgit2 [`git_checkout_tree`](https://libgit2.github.com/libgit2/#HEAD/group/checkout/git_checkout_tree) API actually only supports "a commit, tag or tree".

The following code will reproduce the problem.
```
> node bug.js
[Error: Object to checkout does not match repository]
```
```JavaScript
var git = require("nodegit");
git.Repository.open("myrepo")
.then(function(repo) {
  return repo.getTagByName('v1.0.0').then(function(tag) {
    git.Checkout.tree(repo, tag.id(), { checkoutStrategy: git.Checkout.STRATEGY.SAFE_CREATE })
    .then(function() {
      console.log("ok");
    }).catch(function(err) {
      console.log(err);
    });
  });
});
```

### DIFF
--- a/lib/checkout.js
+++ b/lib/checkout.js
@@ -40,7 +40,7 @@ Checkout.index = function(repo, index, options) {
 *
 * @async
 * @param {Repository} repo
-* @param {Oid|Tree|Commit|Reference} treeish
+* @param {String|Tree|Commit|Reference} treeish
 * @param {CheckoutOptions} [options]
 * @return {Void} checkout complete
 */


srajko:
@rcjsuen Thanks for the doc fix!  It's a little scary that we're letting the `Oid` get all the way to `git_checkout_tree`, which is expecting a `git_object` (and a `git_oid` is not a `git_object`) :-(

`Checkout.tree` does seem to take a `String` though - in your example, you could try `tag.id().toString()`.  Could I bother you to replace `Oid` with `String` in the docs, to reflect that?

rcjsuen:
Thanks for the code review, @srajko! I have updated the `checkout.js` as you requested.

Thanks for offering a solution as well. Though I don't actually need it as this was actually a bug reported by @deowk on Gitter. :P

Issue #489, opened 2015-03-16
HTTP Proxy support


franckstauffer:
It would be great to have support for HTTP/HTTPS proxies for people sitting behind corporate proxies who clone/push/pull to remote repositories. It looks like libgit2 is only supporting this in the winhttp transport backend and that there is an open issue libgit2/libgit2#2555 to support it by default.

johnhaley81:
When that is closed out on libgit2 and released we'll make sure to get that into NodeGit as well. Thanks!

franckstauffer:
As a followup, in the meantime if someone else stumbles upon that problem, using a tool like proxychains-ng rofl0r/proxychains-ng might ease the pain :)

tbranyen:
Closing this as we cannot address it at this time.  As @johnhaley81 said, once it lands and is tagged in libgit2 it'll be available here.

franckstauffer:
Ok fair enough...I only created the issue because @johnhaley81 invited me to do so on gitter!

tbranyen:
Sure, don't take it as a jab, I'm just trying to tidy up the issue tracker :)

johnhaley81:
I just wanted some kind of history :)

maxkorp:
Labeling this, so we can zip through the closed stuff later and see if this has been addressed.

Kerrick:
Looks like libgit2 merged in support for proxies! libgit2/libgit2#3110

johnhaley81:
Working on merging that now :)

vbarshai:
Has this ever been merged? Or is the issue still open? Thanks!

cbargren:
Proxy support did get merged in, this can probably be closed.

vbarshai:
Thanks @cbargren it's been merged in v0.15.0 but looks like the latest builds break:
babel --presets es2015 -d ./dist ./lib (no lib dir)

hellosmithy:
Has anyone managed to get this working behind a corporate proxy?

vbarshai:
@hellosmithy not on our side - ended up having to use proxy exports and native git commands as a workaround until a stable release.

johnhaley81:
@cbargren and @implausible were working on getting something going here but ran into some bigger issues.

cbargren:
[...]

PR #1195, opened 2017-01-24
build against shared library for libcurl

implausible:
To get proxy support into NodeGit, we will need to get libgit configured to use libcurl. To start, we will just try to build nodegit against libcurl using the -lcurl flag.

We will also expose and handle proxy_options, the goal of this PR is to get proxy support in full across all systems.

This should address https://github.com/nodegit/nodegit/issues/489

### DIFF
--- a/generate/input/libgit2-supplement.json
+++ b/generate/input/libgit2-supplement.json
@@ -482,6 +482,10 @@
               "type": "git_fetch_prune_t"
             },
             {
+              "name": "proxy_opts",
+              "type": "git_proxy_options"
+            },
+            {
               "name": "update_fetchhead",
               "type": "int"
             },
--- a/generate/templates/templates/binding.gyp
+++ b/generate/templates/templates/binding.gyp
@@ -88,13 +88,22 @@
               }
             }
           }
-        ], [
+        ],
+        [
+          "OS=='linux' or OS=='mac'", {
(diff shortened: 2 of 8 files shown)


implausible:
Testing this is proving to be quite difficult. Issues are that we want to assert that we (@cbargren in da house) are going over a proxy. We need to figure out a way to create a proxy on the fly on our test boxes that we can tell nodegit to use that also can be checked at the end of the test if it was hit or not... we've looked through several libraries today and so far nothing is working for us :(

implausible:
Getting proxy options fully tested is going to be _very_ difficult. For now, this passes all of the test cases, and does not worsen NodeGit. Had a discussion with @johnhaley81, we'll come back for tests on this when we have a good framework for an on-the-fly forward proxy.

maxkorp:
Probably a custom proxy through express or something would be easiest.

maxkorp:
Looking at this now.

johnhaley81:
Tested this in a proxied environment and confirmed it worked. Thanks @implausible.

saper:
this breaks my FreeBSD build since it cannot find `<curl/curl.h>` (It is in `/usr/local/include/curl/curl.h`).

This change should use `curl-config` to figure out compiler and linking options to get it working.

PR #1199, opened 2017-01-27
Remove Node v0.12 and v5

johnhaley81:
v0.12 is breaking the deploy process now and v5 should be removed in favor of v7 which will come in a later commit.


implausible:
That was fast.

PR #1203, opened 2017-01-28
Specify acceptable types for `lookup` and `dwim` in Reference

rcjsuen:
According to libgit2's APIs, only git_commits and git_tags are acceptable. Update the documentation to accurately reflect this instead of just taking in an `Object`.

https://libgit2.github.com/libgit2/#HEAD/group/reset/git_reset

https://libgit2.github.com/libgit2/#HEAD/group/reset/git_reset_default

### DIFF
--- a/lib/reset.js
+++ b/lib/reset.js
@@ -10,8 +10,8 @@ var _reset = Reset.reset;
  *
  * @async
  * @param {Repository} repo Repository where to perform the reset operation.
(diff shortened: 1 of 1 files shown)


rcjsuen:
JSHint's character restriction per line meant I had to make more changes to the `reset.js` beyond just simply modifying the `@param` tag.

maxkorp:
<3

PR #1205, opened 2017-01-29
Write the index to a repository directly in merge-cleanly.js examples

### LINKED ISSUE #1128: Writing index after merge throws error
When running examples/merge-cleanly.js, the following error is thrown by index.write():
Uncaught Error: Failed to write index: The index is in-memory only
Either index.write() or the example is broken.

rcjsuen:
This fixes #1128. The returned index is only in-memory and doesn't know where to write the .git/index file to so we should just skip the `write()` call and to just only use `writeTreeTo()` instead.


maxkorp:
Excellent find.

PR #1208, opened 2017-01-31
Make waiting for callback result optional

srajko:


srajko:
This change allows callback properties to be given with `waitForResult: false`, which will schedule the callback but not wait for its result.  Instead, the default callback result will be immediately returned to `libgit2`.

To make sure completion callbacks get executed after intermediate callbacks, their two separate queues in `ThreadPool` are now a single queue.

johnhaley81:
Thanks @srajko this also has the nice side effect of making the code easier to read :)

PR #1212, opened 2017-02-01
Replace references of Gitter with Slack in CONTRIBUTING.md

rcjsuen:
It seems like the maintainers of NodeGit are now in the Slack channel instead of Gitter so the `CONTRIBUTING.md` file should be updated accordingly.


johnhaley81:
Thanks @rcjsuen !

PR #1213, opened 2017-02-04
`RevertOptions` parameter in `Revert.revert` is actually optional

rcjsuen:
libgit2's `git_revert` function does not actually require the revert options to be specified. It is merely a [documentation oversight](https://github.com/libgit2/libgit2/pull/4108) on their website. We should flag the `given_opts` argument as being `isOptional` so that `null` can be passed into the JavaScript function.


johnhaley81:
Awesome! Thanks again @rcjsuen :)

PR #1214, opened 2017-02-05
Make `Reset.fromAnnotated` the same as `Reset.reset`

rcjsuen:
The `fromAnnotated` function should be asynchronous in addition to having its `checkout_opts` parameter be optional as the function delegates to the same function as `reset` internally in libgit2.


johnhaley81:
Thanks @rcjsuen!

PR #1215, opened 2017-02-06
String `message` argument of `Stash.save` is optional

rcjsuen:
If you look at libgit2's [stash.h](https://github.com/libgit2/libgit2/blob/d02720d8a653bc564b84c27bfa01a6876bf5bec2/include/git2/stash.h#L59) file, you will see that the `message` parameter is actually optional. We should flag the argument as being optional in NodeGit's JSON file as well.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2175,6 +2175,11 @@
           }
         },
         "git_stash_save": {
+          "args": {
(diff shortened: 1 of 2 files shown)


johnhaley81:
Thanks @rcjsuen for the great PRs!

PR #1218, opened 2017-02-08
build safe convenience wrapper for git_remote_ls

implausible:
Documentation for `git_remote_ls` shows that for it to work, a connection must be established to a remote first. When a subsequent connection is made, retrieved remote heads are destroyed and replaced with a new set.

This manual template locks the remote during execution of 3 actions (git_remote_connect, git_remote_ls, git_remote_disconnect) and copies all the data retrieved from git_remote_ls into memory for nodegit consumption.

The other contributing factor for using a manual template here is git_remote_ls has the only triple pointer use case in libgit2, so writing additional handlers for one case is overkill, especially when considering the volatility of the memory being retrieved. Complications that will exist if there is ever a good reason to build triple pointer handling into our generation system will be handling the bounds of the retrieved double pointer, as there will probably be a corresponding size argument.

### DIFF
--- a/generate/templates/manual/remote/ls.cc
+++ b/generate/templates/manual/remote/ls.cc
@@ -40,42 +40,52 @@ NAN_METHOD(GitRemote::ReferenceList)
 void GitRemote::ReferenceListWorker::Execute()
 {
   giterr_clear();
-  baton->error_code = git_remote_connect(
-    baton->remote,
-    GIT_DIRECTION_FETCH,
-    baton->callbacks,
-    baton->proxy_opts,
-    baton->custom_headers
-  );
 
-  if (baton->error_code != GIT_OK) {
-    baton->error = git_error_dup(giterr_last());
-    delete baton->out;
-    baton->out = NULL;
-    return;
-  }
+  {
+    LockMaster lockMaster(
+      /*asyncAction: */true,
+      baton->remote,
+      baton->callbacks,
+      baton->proxy_opts,
+      baton->custom_headers
+    );
+    baton->error_code = git_remote_connect(
+      baton->remote,
+      GIT_DIRECTION_FETCH,
+      baton->callbacks,
+      baton->proxy_opts,
+      baton->custom_headers
+    );
 
-  const git_remote_head **remote_heads;
-  size_t num_remote_heads;
-  baton->error_code = git_remote_ls(
-    &remote_heads,
-    &num_remote_heads,
-    baton->remote
-  );
+    if (baton->error_code != GIT_OK) {
+      baton->error = git_error_dup(giterr_last());
+      delete baton->out;
+      baton->out = NULL;
+      return;
+    }
 
-  git_remote_disconnect(baton->remote);
-  if (baton->error_code != GIT_OK) {
-    baton->error = git_error_dup(giterr_last());
-    delete baton->out;
-    baton->out = NULL;
-    return;
-  }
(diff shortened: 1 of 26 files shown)


generate/templates/manual/remote/ls.cc
```diff
@@ -0,0 +1,136 @@
...
+
+  {
+    LockMaster lockMaster(
+      /*asyncAction: */true,
+      baton->remote,
+      baton->callbacks,
+      baton->proxy_opts,
+      baton->custom_headers
+    );
+    baton->error_code = git_remote_connect(
```
srajko: Instead of connecting (and disconnecting) as part of `ReferenceList`, it could be just a wrapper over `git_remote_ls`, and that way the user can use it with an already connected remote.

implausible: Done and doner @srajko.

generate/templates/manual/remote/ls.cc
```diff
@@ -0,0 +1,99 @@
...
+    if (baton->error->message)
+    {
+      free((void *)baton->error->message);
+    }
+
+    free((void *)baton->error);
+  }
+  else if (baton->error_code < 0)
+  {
+    Local<v8::Object> err = Nan::Error("Method next has thrown an error.")->ToObject();
```
srajko: "Method next has thrown an error." seems like the wrong error here.

generate/templates/manual/src/functions/copy.cc
```diff
@@ -20,3 +20,15 @@ void git_transfer_progress_dup(git_transfer_progress **out, const git_transfer_p
   *out = (git_transfer_progress *)malloc(sizeof(git_transfer_progress));
   memcpy(*out, arg, sizeof(git_transfer_progress));
 }
+
+git_remote_head *git_remote_head_dup(const git_remote_head *src) {
+  git_remote_head *dest = (git_remote_head *)malloc(sizeof(git_remote_head));
+  dest->local = src->local;
+  git_oid_cpy(&dest->oid, &src->oid);
+  git_oid_cpy(&dest->loid, &src->loid);
+  dest->name = strdup(src->name);
```
srajko: a null-check here like you have below wouldn't hurt

lib/remote.js
```diff
@@ -113,6 +113,15 @@ Remote.prototype.push = function(refSpecs, opts) {
 };
 
 /**
+ * Lists advertised references from a remote
+ *
+ * @async
+ * @return {Promise<Array<RemoteHead>>} a list of the remote heads currently
+ *                                      available on the remote for download
```
srajko: "currently available" might be a little off - isn't it more like what the remote had available when the connection was established?

generate/templates/manual/src/functions/copy.cc
implausible: I think it's always supposed to be defined, but you're right, we should err on the side of safety :man_facepalming:.

danwoodbury:
is there a way to use this but for a specific branch?

Mklueh:
Can sb please tell me how to use this? I can´t find it in the documentation

Just want to get a list of all remote branches from remote without fetching anything like "git ls-remote git@...."

webbrandon:
@Mklueh Looking at the unit test helps :)   Looks like a function on the response:
```
return this.repository.getRemote("origin")
      .then(function(remote) {
        return remote.referenceList();
})
```

PR #1219, opened 2017-02-08
Update merge branches with processMergeMessageCallback

implausible:
When merging branches, we should provide a callback that allows users to inspect and modify the message that will be used for merging.

The generated merge message itself was also incorrect when compared to git core, and so has been updated to match git core.


johnhaley81:
Thanks @implausible!

PR #1220, opened 2017-02-09
Add regression test for libgit2 #4102

rcjsuen:
If `Revwalk.SORT.TOPOLOGICAL` is used in a revwalk, Node will crash. This is caused by [libgit #4102](https://github.com/libgit2/libgit2/issues/4102).

This is [blocking](https://bugs.eclipse.org/bugs/show_bug.cgi?id=511822) the Orion project from upgrading to NodeGit 0.17.0.

### DIFF
--- a/test/tests/revwalk.js
+++ b/test/tests/revwalk.js
@@ -311,6 +311,18 @@ describe("Revwalk", function() {
       });
   });
 
+  it("regression test for libgit2 #4102", function() {
+    this.walker.sorting(NodeGit.Revwalk.SORT.TOPOLOGICAL);
+    this.walker.pushRange("32789a7..c82fb07");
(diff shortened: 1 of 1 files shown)


rcjsuen:
The test is crashing early as expected (since the Revwalk tests are the last one it executes) but the `0` exit code is flagging the builds as green on Travis. Hm...
```
The command "if [ $TARGET_ARCH == "x64" ]; then if [ -z "$TRAVIS_TAG" ] && [ $TRAVIS_OS_NAME == "linux" ] && [ $NODE_VERSION == "6" ]; then npm test && npm run cov && npm run coveralls; else npm test; fi else echo "Not running tests because the binary is not built for 64-bit systems"; fi" exited with 0.
```

PR #1224, opened 2017-02-14
Provide rebase details on finish

implausible:
When using the convenience methods for rebase, we should return all the details about the rebase before finishing. This includes the sha translation map for the rebase operation, and other meta data (old head sha, branch names, new head sha).

This is a breaking API change.

### DIFF
--- a/test/tests/rebase.js
+++ b/test/tests/rebase.js
@@ -1028,22 +1028,18 @@ describe.only("Rebase", function() {
             nextCalls++;
 
             return Promise.resolve();
-          }, function({
-            ontoName,
-            ontoSha,
-            originalHeadName,
-            originalHeadSha,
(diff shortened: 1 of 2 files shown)


implausible:
Destructuring on node4 + coverage tool = broken.

test/tests/rebase.js
```diff
@@ -4,7 +4,7 @@ var local = path.join.bind(path, __dirname);
 var promisify = require("promisify-node");
 var fse = promisify(require("fs-extra"));
 
-describe("Rebase", function() {
+describe.only("Rebase", function() {
```
johnhaley81: This needs to be removed

implausible: This is literally the most important function :+1:

PR #1232, opened 2017-02-24
Use wider int to calculate throttle window

srajko:
Sufficiently large throttle values would cause an overflow - the cast coerces the computation into a sufficiently large int


johnhaley81:
Not totally sure why, but this is killing linux 32-bit.

johnhaley81:
32-bit linux dieing was unrelated and fixed in #1241

PR #1237, opened 2017-02-26
Use curl-config to find curl

saper:
On BSDs headers will be in /usr/local/include which
is not added to the include path automatically.

Follow-up-to: #1195

### DIFF
--- a/generate/templates/templates/binding.gyp
+++ b/generate/templates/templates/binding.gyp
@@ -91,9 +91,9 @@
           }
         ],
         [
-          "OS=='linux' or OS=='mac'", {
+          "OS=='linux' or OS=='mac' or OS.endswith('bsd')", {
             "libraries": [
-              "-lcurl"
+              "<!(curl-config --libs)"
             ]
           }
         ],
--- a/vendor/libgit2.gyp
+++ b/vendor/libgit2.gyp
@@ -293,7 +293,8 @@
         }],
         ["OS=='mac' or OS=='linux' or OS.endswith('bsd')", {
           "cflags": [
-            "-DGIT_CURL"
+            "-DGIT_CURL",
+            "<!(curl-config --cflags)"
           ],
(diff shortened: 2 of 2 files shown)


saper:
Maybe this would even work for Windows?

implausible:
The only reason to bring libcurl into NodeGit is to enable proxy and authenticated proxy transports in OSX and Linux. Windows already has full support for proxies.

implausible:
@saper should rebase this on master so the 32 bit linux tests passes. This is definitely a good idea.

johnhaley81:
@saper could you rebase this please? I'd like to get this in for the next release. Thanks!

saper:
Rebased... well it was master and the time I wrote this :) and there were no conflicts

rsmarples:
What needs to be done here to progress this?
I need these changes for building nodegit on NetBSD.

myfreeweb:
Please merge this, `npm i nodegit` fails on FreeBSD without this, having to locally fix this and `npm link` is getting annoying.

mfechner:
Are there plans to include this patch, I can confirm that it fixed the problem.

antoinell:
The code in this pull request helps with the build on FreeBSD. Hopefully it can get merged soon. I see that there is an error in a build on windows done by appveyor, this seems unrelated to the code changes here which do not affect windows as far as I understand.

Croydon:
@implausible Can we merge this now? Both #1545 and #1549 is containing it too, but getting this done step by step seems reasonable.

implausible:
@Croydon The tests are failing. @saper Any chance you can rebase this to get the tests running again on CI? It looks like Travis doesn't even know about this PR.

Otherwise if @saper is unavailable, I would support it if you, @Croydon, took control of this PR by opening a new branch + PR with it. We would close this PR, reference this PR in your PR, and merge your PR.

PR #1241, opened 2017-02-27
Fix linux 32-bit builds

johnhaley81:


johnhaley81:
The builds were failing because somehow the `sudo: required` broke downloading packages off of `apt-get` (?!?) so removing that makes the build work again. 

`sudo` was required because in order to build NodeGit on 32-bit linux we required a symlink to work around a bug introduced in `gcc-multilib` that renamed `/usr/include/asm` to `/usr/include/asm-generic`. That appears to be fixed so the work-around was no longer needed.

As an extra measure I added back in testing 32-bit builds for linux as well since they were removed originally.

johnhaley81:
image

PR #1244, opened 2017-03-02
Add info about missing libssl-dev

malko:
Add some info in the readme on how to resolve install problem about lifecycleScripts


johnhaley81:
Thanks @malko!

PR #1249, opened 2017-03-10
Use example.com domain for default signature.

novalis:
This domain is guaranteed to be unused (see RFC2606).  It is
inappropriate to use a real domain for a made-up email address.


rcjsuen:
@novalis Thanks for submitting a pull request! Unfortunately, the tests will also have to be updated to reflect this change it seems.

novalis:
Oops, missed those.  Let's see if this works...

rcjsuen:
Looks good and merged! Thanks for the fix, @novalis.

PR #1254, opened 2017-03-22
Make Tree#createUpdated usable

rafaelgieschke:
Currently, `Tree#createUpdated` is not usable as new `TreeUpdate`s cannot be constructed and `createUpdated(repo, nupdates, updates)` only accepts a single `TreeUpdate` as `updates`.

This pull request allows to `new TreeUpdate()` and turns `updates` into an `Array<TreeUpdate>`.

Usage is, e.g., as follows:

```javascript
#!/usr/bin/env node

const Git = require("nodegit");

(async () => {
  const repo = await Git.Repository.init("test", 1);
  const odb = await repo.odb();
  const emptyTree = await repo.getTree(await odb.write("", 0, Git.Object.TYPE.TREE));
  const oid = await odb.write("test\n", 5, Git.Object.TYPE.BLOB);

  const update = new Git.TreeUpdate();
  update.action = Git.Tree.UPDATE.UPSERT;
  update.filemode = Git.TreeEntry.FILEMODE.BLOB;
  update.id = oid;
  update.path = "directory1/file";

  const update2 = new Git.TreeUpdate();
  update2.action = Git.Tree.UPDATE.UPSERT;
  update2.filemode = Git.TreeEntry.FILEMODE.BLOB;
  update2.id = oid;
  update2.path = "directory2/file";

  const treeOid = await emptyTree.createUpdated(repo, 2, [update, update2]);
  console.log(treeOid);
})();
```

PR #1255, opened 2017-03-23
Add Writestream#write

rafaelgieschke:
https://github.com/nodegit/nodegit/pull/1010 added `Blob.createFromStream()` but no way to actually use the resulting `Writestream`.

This pull request adds `Writstream#write`, so you can create blobs like:

```javascript
#!/usr/bin/env node

const Git = require("nodegit");

(async () => {
  const repo = await Git.Repository.init("test", 1);
  const stream = await Git.Blob.createFromStream(repo, "");
  const buffer = new Buffer("test\n");
  await stream.write(buffer, buffer.byteLength);
  const buffer2 = new Buffer("test2\n");
  await stream.write(buffer2, buffer2.byteLength);
  const oid = await Git.Blob.createFromstreamCommit(stream);
  console.log(oid);
})();
```

It does so by declaring a fake `git_writestream_write` function, matching the `write` function in `struct writestream` of libgit2 (<https://github.com/libgit2/libgit2/blob/HEAD/include/git2/types.h#L430>).


johnhaley81:
@rafaelgieschke thanks again. Same as the #1254 with adding tests to ensure functionality.

rafaelgieschke:
Added a test.

implausible:
@rafaelgieschke Any chance you can fix the conflicts? I'd like toget this into the codebase.

rafaelgieschke:
@implausible Thanks for looking at this! I've rebased the PR onto the current master.

zawata:
please rebase and reopen if still relevant

PR #1260, opened 2017-03-27
Reference#rename test

smith-kyle:
Added a test for renaming a ref


rcjsuen:
This test looks good. Thanks for the help, @smith-kyle!

PR #1269, opened 2017-04-12
Add JavaScript version of `git_tag_peel()`

rcjsuen:
The generated code for [`git_tag_peel()`](https://libgit2.github.com/libgit2/#v0.25.1/group/tag/git_tag_peel) is taking a pointer as a to return the peeled object. We should change the argument to be `isReturn` so that the generated API is sane from a JavaScript point of view.


cjhoward92:
Can you rebase this on the current master?

PR #1292, opened 2017-05-27
Fix incorrect api usage, not Time rather TIME

cdosborn:
http://www.nodegit.org/api/revwalk/#SORT


rcjsuen:
I decided to force the merge despite the AppVeyor build failure since it's a change to an example and not any code that is part of the build.

Thanks for the pull request, @cdosborn!

PR #1327, opened 2017-07-13
BREAKING: Added blob methods and converted to async

cjhoward92:
Exposed more blob methods from libgit2. The methods are:

- git_blob_filtered_content
- git_blob_create_fromdisk

Also made the createFromBuffer and createFromWorkdir methods asynchronous.

This is a breaking change to the blob api. createFromBuffer and createFromWorkdir are now asynchronous and thus return promises. This is because we try to maintain a standard of having methods that touch the disk as async to prevent deadlocks. While createFromBuffer does not touch the disk, it has a similar api to the other `createFrom*` methods and we thought it made sense to make it async.

### DIFF
--- a/test/tests/blob.js
+++ b/test/tests/blob.js
@@ -1,16 +1,68 @@
 var assert = require("assert");
 var path = require("path");
 var local = path.join.bind(path, __dirname);
+var promisify = require("promisify-node");
+var fse = promisify("fs-extra");
 
 describe("Blob", function() {
   var NodeGit = require("../../");
 
   var Oid = NodeGit.Oid;
   var Repository = NodeGit.Repository;
+  var Blob = NodeGit.Blob;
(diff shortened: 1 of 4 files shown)


implausible:
Rebase this on master, please.

test/tests/blob.js
```diff
@@ -1,16 +1,80 @@
...
+      var author = signatures[0];
+      var committer = signatures[1];
+
+      return repo.createCommit(
+        "HEAD",
+        author,
+        committer,
+        commitMessage,
+        treeOid,
+        [parent]);
```
implausible: I noticed you have a preference for closing on the same line, can we update the style to open/close on new lines if it's necessary to add new lines?

cjhoward92: Yeah, I will run through my tests and make the changes on Monday.

implausible:
Other than the bblobs this is looking good.

PR #1328, opened 2017-07-16
build: use pkg-config to find curl header

rsmarples:
pkgsrc (as used by NetBSD) installs package headers in non default
locations.
The solution is to use pkg-config to work out the correct location
via CFLAGS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,15 +13,12 @@ sudo: false
 
 env:
   matrix:
-    - export NODE_VERSION="4" TARGET_ARCH="x64"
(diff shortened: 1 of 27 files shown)


tbranyen:
Looks good to me :+1:

rsmarples:
no idea what happened here, I'll redo

myfreeweb:
The `"<!(pkg-config libcurl --cflags 2>/dev/null || echo)"` line is also required on FreeBSD.

rsmarples:
See PR #1237 for a better solution. Bump that one instead ;)

PR #1342, opened 2017-08-02
BREAKING: Convert git_status_file to an async method

cjhoward92:
Convert the NodeGit.Status.file routine to be an async routine. This is to help avoid deadlocks as functions that read from or write to disk should be asynchronous.  This is a breaking change since the function now returns a promise and not the status.


cjhoward92:
Replaced by [#1348](https://github.com/nodegit/nodegit/pull/1348)

Issue #1339, opened 2017-07-28
"examples/add-and-commit.js" throw "Error: Invalid path: xxx" on windows


BoyBoBo:
os: windows 7
nodegit: 0.18.3

This line will throw "Error: Invalid path: xxx"
> index.addByPath(path.join(directoryName,  fileName))

I changed it to the following, it's ok
> index.addByPath(directoryName + '/' + fileName)

rcjsuen:
Thank you for the bug report, @BoyBoBo! I have opened #1346 which fixes the problem.

PR #1346, opened 2017-08-06
Fix example by forcing backslashes with path.posix

rcjsuen:
A user reported in #1339 that an error will occur if you try to run the `examples/add-and-commit.js` example on Windows.
```
$ node add-and-commit.js
nodegit\node_modules\nodegit-promise\lib\done.js:10
      throw err;
      ^

Error: invalid path: 'salad\toast\strangerinastrangeland\theresnowaythisexists\newfile.txt'
    at Error (native)
```
This is because `addByPath` from of `Index` seems to only accept paths with forward slashes. Replacing the [`path.join`](https://github.com/nodegit/nodegit/blob/113493a9d3028a42ff30896ca32791835e8d68d9/examples/add-and-commit.js#L49) with `path.posix.join` will resolve the problem.

### DIFF
--- a/examples/add-and-commit.js
+++ b/examples/add-and-commit.js
@@ -46,7 +46,7 @@ nodegit.Repository.open(path.resolve(__dirname, "../.git"))
 })
 .then(function() {
   // this file is in a subdirectory and can use a relative path
-  return index.addByPath(path.join(directoryName, fileName));
+  return index.addByPath(path.posix.join(directoryName, fileName));
 })
 .then(function() {
   // this will write both files to the index


saper:
Thanks! Can you writer a little test for this one?

rcjsuen:
@saper Sorry, I don't understand. What would be added to the test suite?

rcjsuen:
Someone got hit by this again today. Can we get this merged in so that the example is fixed?

@saper Could you describe what you meant by the addition of a test?

PR #1331, opened 2017-07-25
Adding git_filter support in nodegit

mohseenrm:
Adding support for git_filter to nodegit.

Changes:
Filters are now asynchronous,
FilterRegistry added to register and unregister filters.

PR #1348, opened 2017-08-08
BREAKING: Convert sync methods to async methods

cjhoward92:
Convert the following routines to be async routines:

 - NodeGit.Status.file 
 - NodeGit.Merge.merge
 - rebase.commit
 - rebase.abort
 - Commit.amend
 - Reflog.write

This is to help avoid deadlocks as functions that read from or write to disk should be asynchronous. This is a breaking change since the functions now return a promise and not the original return value.

The addition of filters in [#1331](https://github.com/nodegit/nodegit/pull/1331) caused issues with some core synchronous methods that cause deadlocks, such as merge and disk-reading methods.

This PR replaces [#1342](https://github.com/nodegit/nodegit/pull/1342)

### DIFF
--- a/generate/templates/partials/async_function.cc
+++ b/generate/templates/partials/async_function.cc
@@ -10,11 +10,6 @@ NAN_METHOD({{ cppClassName }}::{{ cppFunctionName }}) {
 
   baton->error_code = GIT_OK;
   baton->error = NULL;
-  {%if cppClassName == "GitStatus" %}
-    {%if cppFunctionName == "File" %}
-  baton->status_flags = (unsigned int *)malloc(sizeof(unsigned int));
-    {%endif%}
-  {%endif%}
 
   {%each args|argsInfo as arg %}
     {%if arg.globalPayload %}
@@ -264,12 +259,6 @@ void {{ cppClassName }}::{{ cppFunctionName }}Worker::HandleOKCallback() {
       {%endif%}
     {%endeach%}
   }
-
-  {%if cppClassName == "GitStatus" %}
-    {%if cppFunctionName == "File" %}
-  free((void *)baton->status_flags);
-    {%endif%}
-  {%endif%}
 
   {%each args|argsInfo as arg %}
     {%if arg.isCppClassStringOrArray %}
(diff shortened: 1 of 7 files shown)


lib/repository.js
```diff
@@ -449,9 +449,13 @@ Repository.prototype.continueRebase = function(
     })
     .then(function(_rebase) {
       rebase = _rebase;
-      return rebase.commit(null, signature);
+      return rebase.commit(null, signature)
+        .catch(function() {
+          // we are ignoring errors. This is to prevent issues caused by making this async
```
rcjsuen: What are these errors that we are supposedly ignoring?

cjhoward92: All of them. The original synchronous method would ignore the libgit2 result and just return the Oid. We ran into issues with this change when rebasing as the `EAPPLIED` error is sometimes thrown when a patch is already applied. Because we switched to the asynchronous pattern the code now throws instead of returning a null Oid. This caused rebase operations to halt when trying to continue rebase in some scenarios.

package.json
```diff
@@ -65,7 +65,7 @@
     "host": "https://nodegit.s3.amazonaws.com/nodegit/nodegit/"
   },
   "scripts": {
-    "babel": "babel --presets es2015 -d ./dist ./lib",
+    "babel": "babel --presets es2015 -d dist lib",
```
cjhoward92: @mohseenrm Why did you make this change?

mohseenrm: for testing purposes, you can discard them

generate/templates/partials/async_function.cc
```diff
@@ -10,6 +10,11 @@ NAN_METHOD({{ cppClassName }}::{{ cppFunctionName }}) {
 
   baton->error_code = GIT_OK;
   baton->error = NULL;
+  {%if cppClassName == "GitStatus" %}
```
implausible: This seems too specific. Check shouldAlloc flag. It should be able to generate this stub without adding this special condition.

generate/templates/partials/async_function.cc
```diff
@@ -260,6 +265,12 @@ void {{ cppClassName }}::{{ cppFunctionName }}Worker::HandleOKCallback() {
     {%endeach%}
   }
 
+  {%if cppClassName == "GitStatus" %}
```
implausible: Same here, we should be able to generate this without a special condition.

PR #1354, opened 2017-08-14
fix issue with re-require()'ing native modules

jmurzy:
Discussion on why this change was required can be found in the following issues:

https://github.com/facebook/jest/issues/3552
https://github.com/facebook/jest/issues/3550
https://github.com/nodejs/node/issues/5016

### DIFF
--- a/generate/templates/templates/nodegit.js
+++ b/generate/templates/templates/nodegit.js
@@ -1,3 +1,4 @@
+var _ = require("lodash");
 var promisify = require("promisify-node");
 var rawApi;
 
@@ -14,6 +15,12 @@ catch (ex) {
 
   rawApi = require("../build/Debug/nodegit.node");
 }
+
+// For disccussion on why `cloneDeep` is required, see:
+// https://github.com/facebook/jest/issues/3552
+// https://github.com/facebook/jest/issues/3550
+// https://github.com/nodejs/node/issues/5016
+rawApi = _.cloneDeep(rawApi);
 
 // Native methods do not return an identifiable function, so we
 // have to override them here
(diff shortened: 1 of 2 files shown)


maxkorp:
What is the additional startup time load on this?

jmurzy:
@maxkorp ~3ms on my MacBook Pro:

```js
const start = performanceNow();
rawApi = _.cloneDeep(rawApi);
const end = performanceNow();
console.log(((end - start) / MILLISECONDS).toFixed(3), 'seconds');
```

```bash
0.003 seconds
```

🍻

jmurzy:
@maxkorp is this PR safe to merge? If so, any rough ETA on when it might land in a release?

sqren:
Are there any blockers, or reasons why this is not merged? Perhaps some tests are needed?

implausible:
I think the addition of a test for this would be grand. I'm looking through the related discussions on what drove this PR and see if we can make it land soon.

jmurzy:
@implausible I'd say existing test cases cover this minor tweak as well. So not sure what else can be tested here. Either way, let me know what you want tested and I'll try to add. Thanks!

jmurzy:
@implausible did you get a chance to review this? Any rough ETA on when it might land in a release? 

Thanks!

jmurzy:
@implausible  did you get a chance to review this? Any rough ETA on when it might land in a release?

Thanks!

axiac:
Jest tests of modules that require `nodegit` still do not work.
Using `--no-cache` (as suggested in Jest issue [#3550](/facebook/jest/issues/3550)) doesn't help.
It works if `--maxWorkers` is set to a value larger than the number of tests but this reaches its limits easily.

Do you have any idea where is the actual problem (NodeGit, Jest or Node)?

PR #1365, opened 2017-09-01
Make compatible with promisify-node 0.4.0

olivierlemasle:
The `test` function is required to force promisify-node to promisify native functions, because async inference based on argument's names cannot work on native functions.

With promisify-node ~0.3.0, functions were always promisify, even if they were not considered asynchronous by async inference. The change of behavior was introduced "insidiously" in https://github.com/nodegit/promisify-node/pull/15.


olivierlemasle:
Superseded by #1519.

PR #1381, opened 2017-10-02
changed wording to be better

TomOrth:


cjhoward92:
I'm not sure I find this wording to be really that much better. I would prefer a reference to the build section as opposed to `Please see the above steps on how to do so`. I would also like the `Note` to be separated out, not in parentheses after the previous statement. 

I would also change the line `Run the following command to run the tests.` to maybe read:

Use `npm test` to run the tests.

No need to have a separate section for a single, two word command.

Also, sorry for the late response. We all appreciate your contribution! Thank you so much for taking the time to do this!

PR #1391, opened 2017-10-16
Updated descriptors for git_index_find and git_index_find_prefix

cjhoward92:
`git_index_find` and `git_index_find_prefix` were not properly defined. Added definitions and tests for both routines, as well as made them asynchronous.


cjhoward92:
@implausible I would like this change in before the next version bump if possible.

PR #1393, opened 2017-10-28
Expose JavaScript function name in a rejected promise

rcjsuen:
I've modified the `async_function.cc` template to include both the names of the JavaScript class and function in the error thrown in a rejected promise. This should hopefully make debugging NodeGit a lot easier when there's a long promise chain.


cjhoward92:
I like this! Nice work. @implausible what do you think? Not a perfect solution but helpful none-the-less.

implausible:
Radical @rcjsuen!

PR #1404, opened 2017-11-26
Generate default options for Revert.revert to prevent an exception

### LINKED ISSUE #1403: (Bug) Revert.revert requires a RevertOptions parameter
Revert.revert requires a Revert Options parameter or a null to be passed into the method unlike other methods such as Merge.merge https://github.com/nodegit/nodegit/blob/6ff67a883c66a1cee23feb53b5f82c36c8a4df04/lib/merge.js#L28-L39
and Revert.commit https://github.com/nodegit/nodegit/blob/6ff67a883c66a1cee23feb53b5f82c36c8a4df04/lib/revert.js#L16-L29. 

When a RevertOptions parameter is not passed into Revert.revert this error is thrown
`(node:46081) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Callback is required and must be a Function.` when talking with @rcjsuen in slack he noticed that the error was misleading and originated from the c++ layer.

Here is the code that threw this error
```
/**
 * @description reverts the git branch to the requested commit
 * @param {String} hash represents the hash of the commit
 * @async
 */
async function revert(hash) {
	return new Promise(async function handlePromise(resolve, reject) {
		try {
			let repo = await openRepository(sitePath);

			logger.info('Setting new head...');
			let newHead = await repo.getCommit(hash);

			logger.info('Reverting...');
			await Git.Revert.revert(
				repo,
				newHead,
			);
			return resolve();
		} catch(err) {
			logger.error(`${err}`);
			return reject(err);
		}
	});
}
```

rcjsuen:
If no options are specified for `Revert.revert`, the promise callback becomes the third parameter for the generated C++ revert.cc file and the (expected) fourth parameter as the promise callback will be unsatisfied which will lead to an exception. To fix this, we should generate a default options for the function instead of the callee doesn't provide one.

This fixes #1403.

### DIFF
--- a/lib/revert.js
+++ b/lib/revert.js
@@ -1,8 +1,10 @@
 var NodeGit = require("../");
+var shallowClone = NodeGit.Utils.shallowClone;
 var normalizeOptions = NodeGit.Utils.normalizeOptions;
 
 var Revert = NodeGit.Revert;
 var _commit = Revert.commit;
+var _revert = Revert.revert;
 
 /**
  * Reverts the given commit against the given "our" commit, producing an index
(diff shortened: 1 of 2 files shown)


lib/revert.js
```diff
@@ -44,3 +45,18 @@ Revert.commit = function(
...
+ * working directory.
+ * 
+ * @async
+ * @param {Repository} repo the repository to perform the revert in
+ * @param {Commit} commit the commit to revert
+ * @param {RevertOptions} revert_options the revert options
+ *                                       (or null for defaults)
+ */
+Revert.revert = function(repo, commit, revertOpts) {
+  revertOpts = normalizeOptions(revertOpts || {}, NodeGit.RevertOptions);
```
implausible: I believe revert options also has checkout and merge options nested in it. You may need to pull those out and normalize them. Checkout the normalizeFetchOptions.js file to see how we've normalized nested options structures before.

rcjsuen:
Thank you for the review, @implausible. Could you take a look at the new changes that I've pushed? Thanks! :)

implausible:
That's awesome, thanks @rcjsuen.

PR #1414, opened 2017-12-15
Exposed filter_list; Fixed discard lines to use filters

cjhoward92:
Exposed the filter_list routines to apply filters to arbitrary files or data. Updated the discardLines routine in repository.js to use filters when they exist. This should create a more reliable output when discarding lines.


cjhoward92:
@implausible If you could give this a look-see next week that would be swell.

generate/templates/manual/src/git_buf_converter.cc
```diff
@@ -1,30 +1,32 @@
...
 
+    result->asize = stringLength;
     result->size = stringLength;
     result->ptr = reinterpret_cast<char *>(memory);
     return result;
   } else {
     return NULL;
   }
-}
+}
```
implausible: Newline please

lib/repository.js
```diff
@@ -729,15 +731,34 @@ Repository.prototype.discardLines =
...
+      if (filterList) {
+        return filterList.applyToFile(repo, filePath);
+      }
 
       return fse.readFile(fullFilePath, "utf8");
     })
     .then(function(content) {
       originalContent = content;
+      if (filterList) {
+        filterList.free();
```
implausible: My first inclination was to express how we shouldn't do this, and that we should strive to improve nodegit's automatic memory management... My second inclination was that we should really throw away the automatic memory management in nodegit and just expose the free methods on anything that needs freeing and leave memory management up to the consumer of nodegit. It's a little more cognitive on the programmer, but it is _way_ more straightforward and potentially complete, compared to what we have now.

cjhoward92: You think I should change it now just to be consistent?

implausible: No. I am pretty certain we're going to move away from our current garbage collection system.

PR #1421, opened 2017-12-26
Annoate Blame.file as async

jmtoball:
cf: https://github.com/nodegit/nodegit/blob/master/test/tests/blame.js#L32


rcjsuen:
Looks good to me! Thanks for the pull request, @jmtoball!

PR #1434, opened 2018-01-26
resolves #1433 TreeEntry#path should return posix path

### LINKED ISSUE #1433: TreeEntry#path should return a posix path
When I use `git ls-tree`, regardless of whether I'm on Linux, macOS, or Windows, it always reports posix paths (paths that use forward slash as the directory separator). This makes sense since git is a database of virtual paths which have nothing to do with the filesystem / platform I'm currently on. nodegit, on the other hand, returns platform-specific paths. In other words, on Windows, the TreeEntry#path returns a different value than on Linux. It's even stranger that I request an entry using a posix path, but I get back a platform-specific path.

Here's a sample of what `git ls-tree -r HEAD` returns on Windows:

```
100644 blob 3fda23c7b550d75eae3d62e471dd35bbc3d6e667    guides/README.md
100644 blob 7a44d0cf690215933974035542d5af6280294982    guides/cloning/README.md
100644 blob 1a9190ab32758755dc1b964e157d2a44abc512d5    guides/cloning/gh-two-factor/README.md
100644 blob d723e52ccf3569b4b25f25ee5d5f21b7a4138453    guides/cloning/gh-two-factor/index.js
100644 blob f6b7c7a3700b1f29bb40b4a9df01795e3a46f106    guides/cloning/index.js
```

Here are the paths nodegit returns on Windows when walking the same tree:

```
guides\README.md
guides\cloning\README.md
guides\cloning\gh-two-factor\README.md
guides\cloning\gh-two-factor\index.js
guides\cloning\index.js
```

This problem can be easily resolved by using [path.posix](https://nodejs.org/api/path.html#path_path_posix) instead of path in tree_entry.js (and perhaps other places).

If the user really wants the Windows-variant of the path, they can easily get it by passing the value to `path.normalize`.

mojavelinux:


mojavelinux:
I'm looking for someone to review this PR and provide me with some feedback.

cjhoward92:
I think this looks pretty good. Was there some issue you were trying to solve with this? Were you having problems? I am curios as to why this is an issue and what, if anything, could be broken by this change (it seems pretty innocuous) .

cjhoward92:
Was `tree_entry.path()` returning broken paths on windows? Something like` C:/path/to\\file\\location`?

mojavelinux:
To start, what nodegit returns differs from what the native git client (`git ls-tree`) returns, even on Windows. On all platforms, paths in a git repository are reported in posix form (using forward slash as the path separator).

```
$ git ls-tree --name-only -r HEAD
README.md
lib/README.md
lib/blame.js
...
test/tests/blame.js
...
```

Where this becomes a problem is when I'm building paths in the TreeWalker. On Windows, I keep having to change backslashes to forward slashes. Otherwise, when I turn around to retrieve a path from the repository, it isn't found. To me, that's where the main inconsistently comes from. If I give a path back to nodegit that nodegit gave to me, it can't find the path in the repository.

There's really no place for backslashes here. Paths in a git repository do not represent real paths (until the part of a checkout). They represent part of the object's identity. Therefore, nodegit shouldn't report different values (and thus break the object's identity) when run on different operating systems.

mojavelinux:
Thanks!

Issue #1318, opened 2017-07-06
File not found on rebase


tjcarroll11:
I'm running a pretty straightforward rebase:

```javascript
return repo.rebaseBranches(
     'refs/heads/test',
     'refs/heads/master',
     null,
     nodegit.Signature.now('Test', 'test@example.com'),
     rebase => {
         console.log("One operation");
         return Promise.resolve();
     },
     rebaseMetadata => {
         console.log("Finished rebase");
         return Promise.resolve();
     }
)
```

Every time I run this, I get the following error

```
{Error: ENOENT: no such file or directory, open '.../.git/rebase-merge/rewritten'
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '.../.git/rebase-merge/rewritten' }
```

My working directory is clean. When I go to my terminal, I can run `git rebase --continue`, and the rebase works with the following message

```
Already applied: 0001 temp
All done.
```

rcjsuen:
@tjcarroll11 Thank you for your bug report. Can you try not providing a `rebaseMetadata` function? Does that help?

tjcarroll11:
Yes. removing that resolves the problem.

rcjsuen:
@tjcarroll11 Okay, you can just extend the promise on the actual rebase instead as a workaround I guess.

```JavaScript
return repo.rebaseBranches(
     'refs/heads/test',
     'refs/heads/master',
     null,
     nodegit.Signature.now('Test', 'test@example.com'),
     rebase => {
         console.log("One operation");
         return Promise.resolve();
     }
))
.then(function() {
  console.log("Finished rebase");
});
```

cjhoward92:
The issue here is that `NodeGit` unfortunately does not check to see if the `rewritten` files exists during the rebase, instead it just assumes the file exists. When we do a fast forward merge, that file is never created and `NodeGit` explodes. Fixed in [#1457](https://github.com/nodegit/nodegit/pull/1457).

PR #1457, opened 2018-03-12
Handle fast-forward merges properly in repository.performRebase

cjhoward92:
Rebases currently expect the `rewritten` file to exist when a rebase occurs and a `beforeFinishFn` is specified. This is not the case when the rebase results in a fast-forward merge as nothing is rewritten.

### DIFF
--- a/lib/repository.js
+++ b/lib/repository.js
@@ -204,9 +204,8 @@ function performRebase(
 ) {
   var beforeNextFnResult;
 
-  /* In the case of FF merges, this will fail
-   * when looking for 'rewritten' so we need
-   * to handle that case.
+  /* In the case of FF merges and a beforeFinishFn, this will fail
+   * when looking for 'rewritten' so we need to handle that case.
    */
   function readRebaseMetadataFile(fileName, determineExistance) {
     const fullPath = path.join(repository.path(), "rebase-merge", fileName);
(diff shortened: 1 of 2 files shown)


cjhoward92:
@implausible

rcjsuen:
Can we get a test added for this?

Will this also fix #1318?

cjhoward92:
I'll try and add a test. Yes, it will fix that issue.

lib/repository.js
```diff
@@ -204,9 +204,17 @@ function performRebase(
 ) {
   var beforeNextFnResult;
 
-  function readRebaseMetadataFile(fileName) {
+  /* In the case of FF merges and a beforeFinishFn, this will fail
+   * when looking for 'rewritten' so we need to handle that case.
+   */
+  function readRebaseMetadataFile(fileName, determineExistance) {
+    const fullPath = path.join(repository.path(), "rebase-merge", fileName);
+    if (determineExistance && !fse.existsSync(fullPath)) {
```
implausible: Let's use the async version of exists.

cjhoward92: Because of how `promisify-node` turns callbacks into promises, `fse.exists` has a bug where you need to call `.catch` to actually get the value you would expect from `.then` so it looks really horrible and is a bit confusing if you ask me. Either that or I can use the callback version, which I don't want to do either. That being said, I don't think the asynchronous version will really gain us anything in this instance so I would rather avoid adding 

```
fse.exists(fillPath).catch(result => /* do work here */).then()
```

to this code block. What do you think?

tbranyen: What bug is this in promisify-node and why can't we fix that there and then improve this code?

cjhoward92: It may not be `promisify-node` as the issue, depending on how you look at it. `exists` has one callback argument. `promisify-node` assumes that first argument is an error and the second argument is the value returned. `exists` doesn't adhere to nodejs callback best practices.

cjhoward92: That being said we can bump `fse` as later versions seem to have fixed this issue, from what I understand, or we can use a different function all together, like `stat` or `access` or something something.

The node documentation recommends we just run a `read` on the file and catch the error, which can also work.

PR #1470, opened 2018-03-23
Add JavaScript Tag.listMatch function for git_tag_list_match

rcjsuen:
I updated the descriptor file to add support for `git_tag_list_match`. This function was asked about on Slack.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2785,6 +2785,19 @@
           },
           "isAsync": true
         },
+        "git_tag_list_match": {
+          "args": {
+            "tag_names": {
+              "isReturn": true,
+              "shouldAlloc": true,
+              "cppClassName": "Array",
+              "jsClassName": "Array",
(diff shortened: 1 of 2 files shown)


sergebesson:
this is me who asked about it on slack. I wanted to use this method to list all tags beginning by "v_". 
Example, following the documentation [nodegit.tag.listMatch](http://www.nodegit.org/api/tag/#listMatch): 
```javascript
const NodeGit = require("nodegit");

NodeGit.Repository.open("MyProject")
  .then((repo) => {
    const listTags = [];
    const result = NodeGit.Tag.listMatch(listTags, "v_*", repo);
    if (result === 0) {
       console.log("list version tags", listTags);
    }
  })
  .catch((error) => {
    console.log("error", error.stack);
  });
```

I saw your changes, is that method going to be async ?

rcjsuen:
@sergebesson Yes, it will be asynchronous.

PR #1498, opened 2018-05-17
Fix return type for Reference.isHead (typed as boolean, but returns a number)

dsabanin:
Both API docs and TypeScript typings say that [Reference.isHead](http://www.nodegit.org/api/reference/#isHead) returns boolean, while in fact it's just a wrapper around Branch.isHead that returns 0 for false, 1 for true. 

This change makes the code do what published API already says it does. Branch.isHead is [correctly typed in API](http://www.nodegit.org/api/branch/#isHead) as returning number.

### DIFF
--- a/lib/reference.js
+++ b/lib/reference.js
@@ -37,7 +37,7 @@ Reference.prototype.isConcrete = function() {
  * @return {bool}
  */
 Reference.prototype.isHead = function() {
-  return Branch.isHead(this);
+  return Branch.isHead(this) === 1;
 };
 
 /**


rcjsuen:
I wonder if it might make more sense to change `Branch.isHead` to return `true` or `false` so that it's more intuitive instead of letting it return `0` or `1`...

dsabanin:
I agree, but I'm not sure what the policy on API changes is. The Branch.isHead is behaving according to the published API, which has it as returning number. Plus some other things on Reference class also have a pattern of returning number 1 as true.

rcjsuen:
@dsabanin For better or worse, we've changed API before. ;) Many functions have changed from synchronous to being asynchronous.

PR #1500, opened 2018-05-18
 Enable git_push_transfer_progress - Help wanted

mitchheddles:
Have enabled the `pushTransferProgress` callback for `RemoteCallbacks`, which is used in `Remote.push`.

The [docs](http://www.nodegit.org/api/push_options/) for `PushOptions` says it accepts callbacks as `RemoteCallbacks` which is correct, but `transferProgress` is only for downloads, libgit has separate [API](https://libgit2.github.com/libgit2/#HEAD/type/git_remote_callbacks) for uploads.

I've been able to get the callback (`pushTransferProgress`) working, similar to `transferProgress` (with throttling). I just need help finishing it off.

**Current Issues**

Generate - The generate callback function uses a regex pattern (`/\s*_(cb|fn)/`) against the function name which isn't satisfied by libgit function name `git_push_transfer_progress`. For now, I've just put in a strict check to bypass.

Testing - I'm unable to get the test case working... I keep getting credentials errors. I was able to get it working by setting up a local bare repo and have the test push to that.

Docs - May need to update the docs to reflect the changes.

It would be great to get some help finishing this off.

Cheers!

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2915,6 +2915,9 @@
     "transfer_progress": {
       "dupFunction": "git_transfer_progress_dup"
     },
+    "push_transfer_progress": {
+      "dupFunction": "git_transfer_progress_dup"
+    },
     "transport": {
       "cType": "git_transport",
       "needsForwardDeclaration": false,
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2311,9 +2311,6 @@
         "push_negotiation": {
           "ignore": true
         },
-        "push_transfer_progress": {
-          "ignore": true
-        },
         "sideband_progress": {
           "ignore": true
         },
@@ -2916,6 +2913,9 @@
       }
     },
     "transfer_progress": {
+      "dupFunction": "git_transfer_progress_dup"
+    },
+    "push_transfer_progress": {
       "dupFunction": "git_transfer_progress_dup"
     },
     "transport": {
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2915,9 +2915,6 @@
     "transfer_progress": {
       "dupFunction": "git_transfer_progress_dup"
     },
-    "push_transfer_progress": {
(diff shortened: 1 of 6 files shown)


rcjsuen:
Perhaps this is something the [`libgit2-supplement.json` file](https://github.com/nodegit/nodegit/blob/04f481c27e30b0306190bad1f0120a962b102cf5/generate/input/libgit2-supplement.json) can help with overriding but I've never added any content to that file myself.

mitchheddles:
@rcjsuen I've got it working using the supplement file and changing `type` to include `_cb`. I'm still not sure if this is the correct approach, but it is working.

I will get onto the test case shortly.

mitchheddles:
@rcjsuen I've setup the test using a local bare repository... Once again, I'm not sure if the approach is correct, but it works.

mitchheddles:
Have rewritten the `isCallbackFunction` to accept a second argument to override the regex test.

This means we can now declare a callback from the [libgit2-supplement.json file](https://github.com/nodegit/nodegit/blob/04f481c27e30b0306190bad1f0120a962b102cf5/generate/input/libgit2-supplement.json) using `isCallback`.

I hope this approach is alright, I can't see another way around it. I'm also not sure why libgit have named the function without `_cb`. 

@rcjsuen Any suggestions for moving this forward?

mitchheddles:
@rcjsuen All checks have passed now, is there any chance of pushing this one through?

generate/input/descriptor.json
```diff
@@ -2918,6 +2915,9 @@
     "transfer_progress": {
       "dupFunction": "git_transfer_progress_dup"
     },
+    "push_transfer_progress": {
```
implausible: I don't believe you need this block. There is no git_push_transfer_progress struct, and it is also theoretically would not compile if that struct was passed to the dupFunction you declared here.

mitchheddles:
@implausible No worries, I've made the changes.

implausible:
Thanks!

ajonp:
Any word on this merging?

PR #1502, opened 2018-05-30
Bump libgit2 to v0.27.1

implausible:


implausible:
Appveyor isn't pulling down the submodule correctly, for what I am assuming is cache related?

PR #1505, opened 2018-06-04
Fix asynchronous testing of signatures

tniessen:
The previous implementation mixed promises and callbacks in a way that was incompatible with mocha. This patch removes the callback and uses promises only. If the test failed, an "unhandled promise rejection warning" was emitted in recent Node.js versions.


tbranyen:
This is much cleaner, thanks!

PR #1509, opened 2018-06-06
Fix behavior of Commit#parent

tniessen:
`Commit#parent` did not assign the `repo` property so subsequent calls which relied on this property (such as `Commit#getTree`) failed. This change works around this problem by imitating what `LookupWrapper` does for `Commit.lookup`, that is, it wraps the actual call in a new function and manually assigns the `repo` property.

This commit also adds a test case for the corrected behavior.

### DIFF
--- a/lib/commit.js
+++ b/lib/commit.js
@@ -17,6 +17,24 @@ var _parent = Commit.prototype.parent;
 Commit.lookup = LookupWrapper(Commit);
 
 /**
+ * @async
+ * @param {Number} n
+ * @return {Commit}
+ */
+Commit.prototype.parent = function(n, callback) {
+  var repo = this.repo;
+  return _parent.call(this, n).then(p => {
+    p.repo = repo;
+
+    if (typeof callback === "function") {
+      callback(null, p);
+    }
+
+    return p;
+  }, callback);
+};
+
+/**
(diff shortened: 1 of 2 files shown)


tniessen:
Ping @tbranyen, @johnhaley81, @maxkorp.

lib/commit.js
```diff
@@ -14,6 +14,30 @@ var _amend = Commit.prototype.amend;
...
+      p.repo = repo;
+
+      if (typeof callback === "function") {
+        callback(null, p);
+      }
+
+      return p;
+    }, callback);
+  };
+})(Commit.prototype.parent);
```
implausible: This is a big change in style from how most of NodeGit is written. I believe we have a pattern for these overrides where we will define:
```javascript
var _parent = Commit.prototype.parent;
```

at the top of the file under _amend.

Then we will reference that later in the file. It's not my favorite scheme, but @maxkorp spent a decent amount of time building a pattern for it that I'm not comfortable tossing out just yet.

tniessen:
@implausible Sorry for the late reply. Is the new commit better?

tniessen:
Ping @implausible, it's been far more than a year, is this still a helpful PR? If not, please close it.

PR #1506, opened 2018-06-05
Bump the promisify-node package

tbranyen:
Fixes Node 10 issues.

PR #1520, opened 2018-07-07
Fixed cannot read property 'name' of null error

sabertazimi:
In `promisify-node~0.3.0`, function `processExports` in `index.js` don't add `null` checking for `exports` argument.It got fixed in `promisify-node~0.5.0`.

Details in [https://github.com/nodegit/promisify-node/issues/24](https://github.com/nodegit/promisify-node/issues/24)

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
     "nan": "^2.10.0",
     "node-gyp": "^3.6.2",
     "node-pre-gyp": "~0.6.39",
-    "promisify-node": "~0.3.0"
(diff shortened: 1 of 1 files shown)


Croydon:
Duplicate of #1506

sabertazimi:
The solutions are the same, and the reasons are different (lol)

tbranyen:
They are the same, we'll track the upgrade in the original PR. Thanks @sabertazimi!

PR #1543, opened 2018-08-29
Update package.json engines to reflect lack of 10 support

maxkorp:
Doesnt mean much directly, except npm will warn and yarn I think will error if you try to install with node 10. Just to provide a bit of feedback to people running into that, since it seems to be coming up a lot.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "lib": "./lib"
   },
   "engines": {
-    "node": ">= 4"
+    "node": ">=4 <10"
   },
   "dependencies": {
     "fs-extra": "~0.27.0",


tbranyen:
I think we're close to getting Node 10 support, so I may want to hold off on this. If I'm wrong, we can merge and push out a patch?

implausible:
We're pretty close. I'm hoping we can wrap it up this week and Node 10 support will arrive at the same time that Node 10 LTS arrives.

maxkorp:
WHAT YOU DON'T WANT TO MERGE THIS NOW THAT IT'S WRONG?!

tbranyen:
I even went so far as to delete you branch, let this be a lesson about being wrong. :trollface:

implausible:
:smile:

PR #1553, opened 2018-09-25
[DO NOT MERGE] Update libssh2 to 1.8.0

Croydon:
Splitting from #1519


Croydon:
Not sure why this fails...

implausible:
We will cover the libssh2 bump in the node 10 support branch, as it's required to get node 10 support.

PR #1555, opened 2018-09-25
Use curl-config to find curl

Croydon:
On BSDs headers will be in /usr/local/include which
is not added to the include path automatically.

This is a rebased version of #1237 
Closes #1237


implausible:
Thanks!

antoinell:
Thanks for having merged this pull request !

PR #1558, opened 2018-10-01
Run submodule updates in sequence rather than parallel

implausible:
Experiencing issues with config locks on electron_npg_automator. This might potentially be the cause.


implausible:
Can't repro the issues for mac locally either. We've never had 2 submodules in nodegit, so this isn't too crazy that it broke electron npg automator a bit.

Issue #1449, opened 2018-02-28
Fix code coverage of tests


Croydon:
Tests weren't run in the CI for some time. Therefore the failure of the tests were hidden.

I changed stuff in #1447 so they would get executed and disabled them temporary (in a visible, easy to switch way) due to the failure.

Eventually they should get fixed and enabled again.

https://travis-ci.org/inexorgame/nodegit/jobs/347256998#L9041

rcjsuen:
> Tests weren't run in the CI for some time. Therefore the failure of the tests were hidden.

I don't understand what you mean, @Croydon. If I look at [nodegit/nodegit's build #3076](https://travis-ci.org/nodegit/nodegit/builds/346476440) for instance, the tests appear to have been run. You can scroll to the bottom of a [log file](https://api.travis-ci.org/v3/job/346476441/log.txt) and see output from Mocha in it.

rcjsuen:
@Croydon Okay. I think what you mean is that the code coverage isn't being run, correct?

JimiC:
Yes.And if someone can tell me what `before_script` I'll be able to sumbit my `travis` config proposal.

Croydon:
No. It fails when it runs.

I have fixed it within my pull requests, that coverages tests get execited when you switch the variable. My last commit disabled coverage tests execution, so we can merge it now for better documentation.

No regression in testing between master and my pull request.
Improvements in documentation, building always for the latest Node minor versions and simplification in the .travis.yml file, visibility when and if coverages tests are getting executed.

Fixing tests are a follow up tasks, therefore this issue.

JimiC:
@Croydon Notice https://github.com/rcjsuen as we discussed that it may be prudent to build docs only on release for the moment.

rcjsuen:
I think the last time code coverage for the C++ files were run was in [Travis CI build #2741](https://travis-ci.org/nodegit/nodegit/builds/195995345). We can see mention of this at the bottom of its [log file](https://api.travis-ci.org/v3/job/195995347/log.txt).
```
...
Processing file src/commit.cc
Processing file src/revert.cc
Processing file src/odb.cc
Processing file src/transfer_progress.cc
Processing file src/signature.cc
Processing file src/proxy_options.cc
Processing file src/clone.cc
Processing file src/index_time.cc
Processing file src/cert_x509.cc
Processing file src/functions/sleep_for_ms.cc
Processing file src/functions/copy.cc
Processing file utils/execPromise.js
[...]

PR #1562, opened 2018-10-02
Turn extended testing back on

stevex86:


Croydon:
This is issue #1449

implausible:
Testing from smart git

stevex86:
Looks like this repo has migrated to GitHub actions. That's awesome! In the interest of keeping this PR from going completely stale, what can we do to get it moving again?

implausible:
My apologies for the reach out. I've been poking at some github APIs. I should've been testing against a private PR.

As far as getting this PR moving again. I dug into extended testing a bit, and it looks like the mechanism for extended testing no longer exists in node-gyp. I don't believe we will be turning extended testing back on without a new plan.

PR #1565, opened 2018-10-08
Generation cleanup

maxkorp:
Changes c++ formatting from astyle to clang-format (using chromium rules) and lints generate folder

### DIFF
--- a/generate/index.js
+++ b/generate/index.js
@@ -10,9 +10,9 @@ module.exports = function generate() {
     .then(function(statuses) {
       var dirtySubmodules = statuses
         .filter(function(status) {
-          return status.onNewCommit
-            || status.needsInitialization
-            || status.workDirDirty;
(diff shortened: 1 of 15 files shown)


maxkorp:
I know where this broke, will get to it later

maxkorp:
Ok I lied, not sure why the new formatting breaks things. Going to omit that commit for now.

LuckyPigeon:
Seems you can't put "+" at the beginning of the line. For Instance, ```return exec(
            "rm "
            + tempSrcDirPath + "/*.cc.orig "
            + tempIncludeDirPath + "/*.h.orig "
          );
```
in generateNativeCode.js, you should change the lines to 
```
return exec(
            "rm " + 
            tempSrcDirPath + "/*.cc.orig " + 
            tempIncludeDirPath + "/*.h.orig "
          );
```

PR #1570, opened 2018-10-17
Garbage collection: Free mostly everything automatically

implausible:
# Garbage Collection in LibGit2
So this is as much as I could glean.

### Reference counted
Most objects that exist in the repository ODB or that extend from object are cached and reference counted. In order to ensure we are cleaning up memory for reference counted libgit2 objects, we should always call the provided libgit2 free method on the pointers when v8 garbage collects them. A lot of these reference counted objects have an underlying memory dependency with the repository, so the repository pointer will need to be kept around until the last call to the last object related to the repository has been freed. This means that most reference counted objects are owned by the repository. This class of objects will always call the free function, even if they have an owner.

### Borrowed pointers
Borrowed pointers are like fields on a struct. They should never be freed. Instead, we should build an owner dependency on the v8 object that contains the struct that this borrowed pointer came from. This ownership dependency will keep the memory alive for as long as the borrowed pointer is alive. 

### Borrowed, duplicable pointers
Certain borrowed pointers (git_oid, git_signature...) are easily duplicable and may have provided duplication functions from libgit2. They may also be easily copied using memcpy. If a pointer is borrowed, but duplicable, no ownership relationship should be established. Instead, we should duplicate the memory and always free when v8 garbage collects us.

### Non reference counted, non duplicable, mega owners
This mostly applies to the repository. The repository has a huge network of underlying relationships between the moment you git_repository_open. Almost every reference counted object has an underlying relationship with the repository via an odb cache or just plain having a field pointing back to the repository. Most reference counted objects will form an owned relationship with the repository. Unfortunately, there are several ways to get back to the same repository pointer that you got when you performed git_repository_open. That means that we need to keep track of how many times we've seen the repository. We now have an external reference counting mechanism to keep track of pointers like the repository. As new v8 objects get created that wrap the repository pointer, we reference count. When every v8 object that wraps the repository has been garbage collected, the last one to be collected will also perform the free method on the repository.
I've currently flagged these as isSingleton and refer to them as such, but that is a silly name and I am open to suggestions (if I don't find a better name myself :smile:).

### OwnerFn
I've added a new field to the ownerFn in the descriptor.json. An example of the ownerFn is `git_commit_owner`. You pass it a commit pointer and it will give you the repository pointer that it is bound to. We now utilize these functions to generate a persistent handle to the repository. When a new commit is created, we bind it to the repository via git_commit_owner. As mentioned above, we will create a new v8 object wrapper around the repository handle and reference count it. This will at least guarantee that even though it's a different v8 object wrapper that we persist, it's still persisting the important underlying libgit2 pointer.

### ownedBy
You can now specify ownedBy fields in arguments to instruct the descriptor that a particular argument from JS is the owner of a return result. This extends ownedByThis behavior by being more open to other ownership configurations than a this relationship.

## Need to do still
- [x] Walk through documentation and ensure ownedBy is set on arguments at the appropriate place
- [ ] Callback memory management is still set to never free callback parameters. We will need to find out how to fish ownership details via the async payload. I am not sure if I want to complete this in this PR or push it into the future, as I imagine callbacks are a much smaller memory foot print than being able to free the repository.
- [ ] Write additional segfault tests like the ones seen in revwalk for the config, which is borderline similar to the repository, but not quite as complex.
- [ ] Write additional segfault tests in general utilizing the GC.

## Future work
- Maybe callback memory managent, depends on if we can get something simple done in this PR.
- Complex owner scenarios like git_filter. It would be nice if instead of having to write a full template, we would be capable of patching the HandleOKCallback with a custom owner resolver.
- It would be nice if we also had an overwrite block specifically targeting function calls in AsyncWorker::Execute. For an example of what I am referring to, check out generate/templates/manual/clone/clone.cc.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -76,10 +76,34 @@
   "types":
   {
     "annotated_commit": {
+      "selfFreeing": true,
       "functions": {
+        "git_annotated_commit_free": {
+          "ignore": true
+        },
         "git_annotated_commit_id": {
           "return": {
             "ownedByThis": true
+          }
+        },
+        "git_annotated_commit_from_ref": {
+          "return": {
+            "ownedBy": "repo"
+          }
+        },
+        "git_annotated_commit_from_fetchhead": {
+          "return": {
+            "ownedBy": "repo"
+          }
+        },
+        "git_annotated_commit_lookup": {
+          "return": {
+            "ownedBy": "repo"
+          }
+        },
+        "git_annotated_commit_from_revspec": {
+          "return": {
+            "ownedBy": "repo"
           }
         }
       }
@@ -115,13 +139,37 @@
       }
     },
     "blame": {
+      "selfFreeing": true,
       "cType": "git_blame",
       "functions": {
+        "git_blame_buffer": {
+          "args": {
+            "out": {
+              "ownedBy": "reference"
+            }
+          }
+        },
         "git_blame_file": {
           "args": {
+            "out": {
+              "ownedBy": "repo"
+            },
             "options": {
               "isOptional": true
             }
+          }
+        },
+        "git_blame_free": {
+          "ignore": true
+        },
+        "git_blame_get_hunk_byindex": {
+          "return": {
+            "ownedByThis": true
+          }
+        },
+        "git_blame_get_hunk_byline": {
+          "return": {
+            "ownedByThis": true
           }
         }
       }
@@ -135,6 +183,10 @@
     },
     "blob": {
       "selfFreeing": true,
+      "ownerFn": {
+        "name": "git_blob_owner",
+        "singletonCppClassName": "GitRepository"
+      },
       "functions": {
         "git_blob_create_frombuffer": {
           "isAsync": true,
@@ -176,6 +228,9 @@
         "git_blob_create_fromchunks": {
           "ignore": true
         },
+        "git_blob_dup": {
+          "ignore": true
+        },
         "git_blob_filtered_content": {
           "isAsync": true,
           "isPrototypeMethod": false,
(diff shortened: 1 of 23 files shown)


implausible:
This tests really well on linux locally, but I kinda want to see if it's stable across platforms.

implausible:
@srajko If you're out there, I will have a more comprehensive write up to how I've extended your work, tomorrow.

implausible:
@maxkorp @tbranyen @srajko I've updated this with a big ol' write up describing what I'm doing.

@ethomson @carlosmn If either of you could fact check me on my GC write up, that would be beneficial.

implausible:
Also, It looks like osx is failing on one test, but because it can't find unordered_map in node 6. @stevex86 suggests there is a flag I can utilize to imagine that into existence.

implausible:
I don't have OSX available, so @stevex86 is gonna help me fix the unordered_map issue.

implausible:
Looks like we will need to revisit git_buf as it really does not behave the same as other git_*_free calls. See https://github.com/libgit2/libgit2/blob/master/include/git2/buffer.h#L22

stevex86:
I got the Node 6 target to build on MacOS. This does change minimum support version of MacOS to from 10.7 (circa 2011) to 10.9 (circa 2013).

tbranyen:
👏 I'll dig in more today. One question now though, is if we should opt folks in, or if certain tasks are better to use free manually? If there is no case where calling free is better that automatic gc, we should probably think of removing those methods from being exported.

implausible:
I've actually removed the capacity for free functions to be called. It's all automatic if this PR lands.

maxkorp:
So, I haven't experimented with it too thoroughly, but I've looked through it a bunch and it *looks* good.

maxkorp:
Yeah, I'm good with landing this once we cut 0.23

PR #1574, opened 2018-10-25
Clean up libssh2 configure step

stevex86:
Fixes some lingering cruft from #1569

### DIFF
--- a/generate/templates/templates/binding.gyp
+++ b/generate/templates/templates/binding.gyp
@@ -5,12 +5,10 @@
         ["OS=='win'", {
           "variables": {
             "is_electron%": "1",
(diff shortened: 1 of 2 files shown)


generate/templates/templates/binding.gyp
```diff
@@ -5,12 +5,10 @@
         ["OS=='win'", {
           "variables": {
             "is_electron%": "1",
-            "openssl_include_dir%": "<(module_root_dir)\\vendor\\openssl"
```
implausible: Since we're cleaning up cruft, there's no reason to have a condition on `OS=='win'` anymore. We should remove that and just set `is_electron%` to `1` or `0`.

PR #1583, opened 2018-11-08
Add support for directories when using `fileHistoryWalk`

spacedragon:


emilsedgh:
I was just looking at the documentation to see how this could be done. Any updates on merging this?

generate/templates/manual/revwalk/file_history_walk.cc
```diff
@@ -185,7 +185,7 @@ void GitRevwalk::FileHistoryWalkWorker::Execute()
 
         if (isEqualNewFile) {
           std::pair<git_commit *, std::pair<char *, git_delta_t> > *historyEntry;
-          if (!isEqualOldFile) {
+          if (!isEqualOldFile || delta ->status == GIT_DELTA_RENAMED) {
```
implausible: Can you fix the spacing on the arrow between delta and status?

Also why was adding git_delta_renamed here necessary?

spacedragon: Before this pr, when:
1. `isEqualNewFile == true` && `isEqualOldFile == false`  //  the file  is created or a file  is renamed
2. `isEqualNewFile == true` && `isEqualOldFile == true` //  the file content changed
...

After this pr,`strncmp` changes the something,    
1. `isEqualNewFile == true` && `isEqualOldFile == false`  //  the file  is created in this folder or a file  is moved to this folder
2. `isEqualNewFile == true` && `isEqualOldFile == true` //  the file content changed or 
**a file is renamed but stays in the same folder**

Without `delta ->status == GIT_DELTA_RENAMED`, the **'a file is renamed but stays in the same folder'** will be handled line 194, which will cause panic when we collect the results here at line  [266](https://github.com/nodegit/nodegit/pull/1583/files#diff-c6f6beacdd90ab36e157ae467f8d7e5cR266)

PR #1586, opened 2018-11-19
Fix typo in GitHub issue template

foolip:


rcjsuen:
Thanks for catching this, @foolip!

PR #1598, opened 2018-12-11
hosting binary files in s3

spacedragon:


spacedragon:
This was not supposed to send here, sorry.

Issue #1596, opened 2018-12-09
Strange C++ Error probably on Reset


GCSBOSS:
## System information
  - node version: v10.8.0.
  - npm or yarn version: 6.3.0
  - OS/version/architecture: Windows 10 x64 (10.0.17134)
  - Applicable nodegit version: 0.23.0

![image](https://user-images.githubusercontent.com/11818414/49699293-6ad89a80-fbb6-11e8-91e5-1bb4b3524c75.png)

`npm ERR! Exit status 3221225477`

Relevant code: 

```js
let current = await repo.getHeadCommit();
await repo.fetchAll();
let newest = repo.getReferenceCommit('origin/master');
if(current.id == newest.id)
    return;
await Reset(repo, newest, Reset.TYPE.HARD);
```

GCSBOSS:
This fixed the code, though that error should probably be handled more gracefully.
```js
let current = await repo.getHeadCommit();
await repo.fetchAll();
let newest = await repo.getReferenceCommit('origin/master');
if(current.sha() == newest.sha())
    return;
await Reset(repo, newest, Reset.TYPE.HARD);
```

implausible:
We should also be very careful when reading from a repo and writing to a repo at the same time.

implausible:
It's possible we could improve the errors; however, in general passing a promise into a method that doesn't expect it is going to lead to a crash.

Issue #1127, opened 2016-09-15
Tighter type checking on passed in nodegit methods 


maxkorp:
As seen here: https://github.com/nodegit/nodegit/issues/1120

Generated code is as such
```
NAN_METHOD(GitTag::List)
{

  if (info.Length() == 0 || !info[0]->IsObject()) {
    return Nan::ThrowError("Repository repo is required.");
  }
```

Where info[0] is meant to be a GitRepo object. Templates need to be fixed up to support this.

rcjsuen:
Can close #1070 and #1183 as duplicates of this I think.

johnhaley81:
@rcjsuen thanks. Closed those other issues in favor of this one. Updates will be posted here.

rcjsuen:
I think we want to use the [`HasInstance(*)`](https://v8docs.nodesource.com/node-7.4/d8/d83/classv8_1_1_function_template.html#a0087e655199fc379268786674f03bc3f) function of a v8 `FunctionTemplate` though I'm not really quite sure how we can randomly access a prototype constructor's `FunctionTemplate` instance. At least, I think that's what we would need to implement this...? o_o

PR #1603, opened 2018-12-18
Check parameters before performing reset

rcjsuen:
If a promise is passed down to the native C++ layer, v8 will crash. To ensure that this does not happen, a check will be performed in the JavaScript layer to verify that the repository is the same as the repository of the commit object.

While logically the commit of one repository instance is the same as a commit of another repository instance (if they point to the same `.git` folder), programmatically they are not equivalent as a check is done
on libgit2's side to ensure that they come from the same thing. As a result, the check on the JavaScript side purely compares the pointers over some other logical piece of information (such as its path or working directory).

This is not a general fix for #1127 but is simply meant to help the case of #1596. I have not put much thought into it but perhaps the generator could be changed so that at least `Repository` and `Commit` objects could be verified?

### DIFF
--- a/lib/reset.js
+++ b/lib/reset.js
@@ -45,7 +45,11 @@ Reset.default = function(repo, target, pathspecs) {
  */
 Reset.reset = function(repo, target, resetType, opts) {
   opts = normalizeOptions(opts, NodeGit.CheckoutOptions);
-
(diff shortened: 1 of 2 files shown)


rcjsuen:
Thanks for reviewing and merging, @implausible!

Issue #1597, opened 2018-12-10
Unable to revert merge commit because RevertOptions cannot be passed in NodeGit


kevin-tla:
I am trying to revert a branch commit by using the Revert.revert() and trying to pass in the RevertOptions, but I am unable to create that instance to give it the necessary values. I get a  "Property 'RevertOptions' does not exist on type 'typeof'...".

I am using Typescript. I wrote it as:

const revertOpts = new NodeGit.RevertOptions(); 

and get that error, however I have no problems when creating an instance of MergeOptions. Also noticed that RevertOptions is also an interface while MergeOptions is a class

Am I doing this correctly?

implausible:
Construct your options as a plain object. The object will magically convert into revert options along the way.

kevin-tla:
> Construct your options as a plain object. The object will magically convert into revert options along the way.

I tried what you recommended: 

NodeGit.Revert.revert(gitRepo, commitid, {mainline: 1});  

And I got a Segmentation Fault. Passed in the object with all the fields too 

{checkoutOpts: null, mainline: 1, mergeOpts: null, version: null} , and get the same error.

implausible:
I was able to reproduce this. Seems there was a slight error in Revert.revert on line 72 of revert.js

implausible:
https://github.com/nodegit/nodegit/pull/1605 will solve the issue. Thanks for bringing this to my attention.

PR #1605, opened 2018-12-20
Fix segfault in NodeGit.Revert.revert

implausible:
Seems we weren't normalizing mergeOpts correctly. Closes https://github.com/nodegit/nodegit/issues/1597


rcjsuen:
I was able to reproduce the crash and the suggested fix resolves the problem. We should add a test if possible (here or in a separate change).

PR #1612, opened 2019-01-23
Add yarn install

LuckyPigeon:


lifecycleScripts/preinstall.js
```diff
@@ -9,22 +9,46 @@ module.exports = function prepareForBuild() {
...
+      var generate = require(local("../generate"));
+      return submodules()
+        .then(function() {
+          return generate();
+        });
+    }
+    })
+    .catch(function(err) {
+      console.log("npm install failed, change to yarn install");
+      return exec("yarn --version")
```
implausible: It might be better to move this particular block to immediately after running npm install, so you only have to fall into the submodules step once.

LuckyPigeon:
@implausible I've seen the change requested. I not entirely understand what your "fix the indentation" means, could you explain it more precisely?

lifecycleScripts/preinstall.js
```diff
@@ -9,12 +9,26 @@ module.exports = function prepareForBuild() {
 
   return exec("npm -v")
     .then(function(npmVersion) {
-      if (npmVersion.split(".")[0] < 3) {
-        console.log("[nodegit] npm@2 installed, pre-loading required packages");
-        return exec("npm install --ignore-scripts");
-      }
+    if (npmVersion.split(".")[0] < 3) {
```
implausible: The indentation here has gone from 6 spaces to 4 spaces.

lifecycleScripts/preinstall.js
```diff
@@ -9,12 +9,26 @@ module.exports = function prepareForBuild() {
...
+      console.log("[nodegit] npm@2 installed, " + 
+                  "pre-loading required packages");
+      return exec("npm install --ignore-scripts");
+    }
+
+    return Promise.resolve();
```
implausible: This return lines up with the indentation of its closing brace.

Would you kindly update the indentation from lines 12 to 18 to be at least 6?

LuckyPigeon: Updated!

LuckyPigeon:
@implausible Updated

implausible:
Hrmm.. So upon further review of this code, I noticed that the current patch is proposing to run `yarn install --ignore-scripts`, but I think this is actually irrelevant.

If you follow the code, we perform `npm -v` to check if the version of npm is less than 2, and _then_ we run `npm install --ignore-scripts`. For all versions `npm@3` and above, we don't run an install script.

I believe that yarn's behavior will mirror that of `npm@3`'s behavior, and we shouldn't need to run `yarn install --ignore-scripts` at all.

I think the patch that we're actually looking for looks like this:
https://github.com/implausible/nodegit/commit/4a1aaf7cee0b5e3f556c03367c11b440059f6497

implausible:
Close in favor of https://github.com/nodegit/nodegit/pull/1644

PR #1616, opened 2019-01-28
Add Commit#amendWithSignature

stevex86:
This change adds signing capabilities for amending commits.

### DIFF
--- a/lib/commit.js
+++ b/lib/commit.js
@@ -54,13 +54,13 @@ Commit.prototype.amend = function (
 /**
  * Amend a commit with the given signature
  * @async
- * @param {String} update_ref
+ * @param {String} updateRef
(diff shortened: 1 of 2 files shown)


lib/commit.js
```diff
@@ -50,6 +51,135 @@ Commit.prototype.amend = function (
...
+    var commitContent;
+    var commitOid;
+    var commit;
+
+    var createCommitPromise = Promise.all(promises)
+      .then(function(results) {
+        treeObject = results[0];
+
+        parents = [];
+        for (var i = 0; i < parentOids.length; i++) {
```
implausible: if you're going to use fp in this method, we should use fp here, as well.
`parents = fp.tail(results);`

PR #1617, opened 2019-01-29
Fix issues with Commit#amendWithSignature

stevex86:
Fixes two issues with the `Commit#amendWithSignature` method:
1. Non-null arguments which override the values from the commit object were being ignored.
2. An extra newline was being inserted into the commit message. This caused extraneous newlines to be appended to commit messages, esp. when the amend operation was repeated.

### DIFF
--- a/lib/commit.js
+++ b/lib/commit.js
@@ -146,7 +146,7 @@ Commit.prototype.amendWithSignature = function(
         );
       })
       .then(function(commitContentResult) {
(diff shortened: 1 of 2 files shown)


stevex86:
As a follow up, I would love to see if we could use the `git_message_prettify` function in libgit2 for formatting messages before passing them into the commit buffer. However, it does not seem to be exposed at the moment.

PR #1621, opened 2019-02-05
Use same API for signingCb in all places that can be crypto signed

implausible:
Any place where we implement a callback pattern for signing commits/tags follows the same API:

`type SigningCB = (content: string) => { code: number, field?: string, signedData?: string };`

What still needs to be done:
- [x] Add new test cases for amend
- [x] Add new test cases for commit
- [x] Add new test cases for rebase
- [x] Add new test cases for tag

### DIFF
--- a/test/tests/commit.js
+++ b/test/tests/commit.js
@@ -450,45 +450,45 @@ describe("Commit", function() {
     });
   });
 
+  describe("amendWithSignature", function() {
+    it("can amend with signature", function() {
+      const signedData = "-----BEGIN PGP SIGNATURE-----\n" +
+      "\n" +
+      "iQJHBAEBCAAxFiEEKdxGpJ93wnkLaBKfURjJKedOfEMFAlxPKUYTHHN0ZXZla0Bh\n" +
+      "eG9zb2Z0LmNvbQAKCRBRGMkp5058Q3vcD/0Uf6P68g98Kbvsgjg/aidM1ujruXaw\n" +
+      "X5WSsCAw+wWGICOj0n+KBnmQruI4HSFz3zykEshuOpcBv1X/+huwDeB/hBqonCU8\n" +
+      "QdexCdWR70YbT1bufesUwV9v1qwE4WOmFxWXgwh55K0wDRkc0u2aLcwrJkIEEVfs\n" +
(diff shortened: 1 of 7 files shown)


test/tests/commit.js
```diff
@@ -450,45 +450,45 @@ describe("Commit", function() {
...
 
-    return NodeGit.Repository.open(reposPath)
+      return NodeGit.Repository.open(reposPath)
       .then(function(repoResult) {
         repo = repoResult;
         return repo.getHeadCommit();
       })
       .then(function(headCommit) {
-        message = headCommit.message();
+        message = headCommit.message() + "\n";
```
stevex86: The messages should be the same here. How come we have to append a `\n` now?

test/tests/commit.js
```diff
@@ -511,65 +510,65 @@ describe("Commit", function() {
...
```
stevex86: Ditto

test/tests/rebase.js
```diff
@@ -1725,4 +1725,355 @@ describe("Rebase", function() {
...
+        // Make sure we don't crash calling the signature CB
+        // after collecting garbage.
+        garbageCollect();
+
+        return rebase.commit(null, ourSignature);
+      })
+      .then(function() {
+        assert.fail("rebase.commit should have failed");
+      }, function(error) {
+        if (error && error.errno === -1) {
```
stevex86: Should be `NodeGit.Error.CODE.ERROR`

test/tests/commit.js
implausible: So when I moved the test into the new describe block, the tests were reordered. Now when the tests run, the previous commit's message does not have a new line, and so amendWithSignature appends a newline to the message. This is more of a test fixture issue, as that newline is required now that we sign the commit content.

dabutvin:
FWIW I just verified the changes here in commit.js still lead to a verified commit signature in the github ui using 

```js
repo.createCommitWithSignature(
    'HEAD',
    author,
    committer,
    commitMessage,
    oid,
    [parent],
    'gpgsig',
    onSignature
  )

...

function onSignature(tosign) { ... }
```

image

implausible:
@dabutvin Sorry to shift this around from underneath ya :smile:. We are hoping to stabilize by release of v0.25.0.

dabutvin:
no worries, the update to take the head parameter is a huge improvement for using this.

PR #1624, opened 2019-02-07
Fix non-existent / dangling refs cause Repository.prototype.createCommitWithSignature to fail

stevex86:
`Repository.prototype.createCommit` is able to handle updates to refs which do not exist because of the internal implementation of `git_commit_create` in libgit2. This PR exposes a reimplementation of this functionality so that `Repository.prototype.createCommitWithSignature` exhibits the same functionality.

### DIFF
--- a/lib/reference.js
+++ b/lib/reference.js
@@ -163,9 +163,12 @@ Reference.updateTerminal = function (
     })
     .then(() => NodeGit.Reflog.read(repo, refName))
     .then((reflog) => {
-      // We may want some kind of transactional logic for this
-      // There is a theoretical timing issue that could result in updating
-      // the wrong reflog
+      // Janky, but works. Ideally, we would want to generate the correct reflog
+      // entry in the first place, rather than drop the most recent entry and
+      // write the correct one.
+      // NOTE: There is a theoretical race condition that could happen here.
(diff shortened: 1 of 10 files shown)


lib/repository.js
```diff
@@ -884,13 +900,16 @@ Repository.prototype.createTag = function(oid, name, message, callback) {
...
  */
 Repository.prototype.defaultSignature = function() {
-  var result = NodeGit.Signature.default(this);
-
-  if (!result || !result.name()) {
-    result = NodeGit.Signature.now("unknown", "unknown@example.com");
-  }
-
-  return result;
+  return NodeGit.Signature.default(this)
```
implausible: Can you update the JSDoc above this to be `@async`

lib/reference.js
```diff
@@ -63,3 +63,114 @@ Reference.prototype.isValid = function() {
...
+  const { email, name } = repo.ident();
+  if (email && name) {
+    return Promise.resolve(NodeGit.Signature.now(name, email));
+  }
+
+  return NodeGit.Signature.default(repo)
+    .catch(() => NodeGit.Signature.now("unknown", "unknown"));
+};
+
+Reference.updateTerminal = function (
```
implausible: Can you JSDoc this?

PR #1634, opened 2019-02-19
Add support for building on IBM i (PASE)

DavidRusso:
These changes allow nodegit to build in the IBM i operating system's PASE environment.

https://en.wikipedia.org/wiki/IBM_i

### DIFF
--- a/generate/templates/templates/binding.gyp
+++ b/generate/templates/templates/binding.gyp
@@ -1,6 +1,7 @@
 {
   "variables": {
-    "is_electron%": "<!(node ./utils/isBuildingForElectron.js <(node_root_dir))"
+    "is_electron%": "<!(node ./utils/isBuildingForElectron.js <(node_root_dir))",
+    "is_IBMi%": "<!(node ./utils/isBuildingForIBMi.js)"
   },
 
   "targets": [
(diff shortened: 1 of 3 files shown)


generate/templates/templates/binding.gyp
```diff
@@ -1,6 +1,7 @@
 {
   "variables": {
-    "is_electron%": "<!(node ./utils/isBuildingForElectron.js <(node_root_dir))"
+    "is_electron%": "<!(node ./utils/isBuildingForElectron.js <(node_root_dir))",
+    "is_IBMi%": "<!(node ./utils/isBuildingForIBMi.js)"
```
implausible: `"is_IBMi%": "OS == 'aix' and <!(node -e "console.log(os.type())) == 'OS400'"`

vendor/libgit2.gyp
```diff
@@ -8,7 +8,8 @@
     "openssl_enable_asm%": 0, # only supported with the Visual Studio 2012 (VC11) toolchain.
     "gcc_version%": 0,
     "is_electron%": "<!(node ../utils/isBuildingForElectron.js <(node_root_dir))",
-    "is_clang%": 0
+    "is_clang%": 0,
+    "is_IBMi%": "<!(node ../utils/isBuildingForIBMi.js)"
```
implausible: `"is_IBMi%": "OS == 'aix' and <!(node -e "console.log(os.type())) == 'OS400'"`

generate/templates/templates/binding.gyp
DavidRusso: That code doesn't seem to work. It seems to assign the string "OS == 'aix' and OS400 == 'OS400' to the variable. The simplest way I found to do it is like this:

`is_IBMi%": "<!(node -p \"os.platform() == 'aix' && os.type() == 'OS400' ? 1 : 0\")"`

Would that be acceptable?

DavidRusso: Also, no, unfortunately I'm not aware of any CI tools for test / build on IBM i.

implausible: Yes that would be acceptable.

PR #1664, opened 2019-04-17
Support signing in Repository#mergeBranches

ianhattendorf:


lib/repository.js
```diff
@@ -1695,13 +1696,19 @@ Repository.prototype.mergeBranches = function(
...
   mergeOptions,
-  processMergeMessageCallback
+  mergeBranchOptions
 ) {
   const repo = this;
   let fromBranch;
   let toBranch;
-  processMergeMessageCallback = processMergeMessageCallback ||
+  // Support old parameter `processMergeMessageCallback`
+  const isOldOptionParameter = typeof mergeBranchOptions === "function";
```
implausible: Can you output a console.error warning that this parameter is deprecated and should be updated to the new object type?

implausible: Any tests that use this method should also be updated to use the new parameter format.

Issue #1661, opened 2019-04-16
yarn audit failure


Toxicable:
yarn audit
nodegit@0.25.0-alpha.9
```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Arbitrary File Overwrite                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ tar                                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=4.4.2                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ nodegit                                                      │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ nodegit > node-gyp > tar                                     │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://www.npmjs.com/advisories/803                         │
└───────────────┴──────────────────────────────────────────────────────────────┘
```
related https://github.com/nodejs/node-gyp/issues/1717

Toxicable:
Looks like this is fixed in v4.0.0
Could we upgrade this dep to fix this issue?

PR #1672, opened 2019-05-03
Dependency upgrade node-gyp upgraded to 4.0.0

davidsarkany:
#1661 Arbitrary File Overwrite vulnerability fixed


implausible:
Thanks! I'm intending to have a new version out today.

PR #1676, opened 2019-05-16
Implement faster file history walk

ianhattendorf:


generate/templates/manual/revwalk/file_history_walk.cc
```diff
@@ -34,251 +214,222 @@ NAN_METHOD(GitRevwalk::FileHistoryWalk)
...
       }
-    }
 
-    git_diff_free(diffs);
+      delete fileHistoryEvent;
 
-    if (!flag && nextCommit != NULL) {
-      git_commit_free(nextCommit);
+      if (firstMatchingParentIndex.first) {
+        git_commit_free(parentCommit);
```
implausible: These need to be freed at the same time we delete fileHistoryEvent or else we're leaking memory on non-matching on merge commits

PR #1696, opened 2019-07-10
Implement support for Node 12

### LINKED ISSUE #1680: Need to implement support for node 12
We need to go through the deprecations list in the compilation phase of our node 10 builds and fix any and all deprecations so that we can update to node 12.

tiggerite:
Fixes #1680 - tested on Node 10.16.0 and Node 12.6.0


implausible:
Hi there! Thanks for working on this! Can you modify the travis.yml and appveyor.yml files to specify tests on node 12?

tiggerite:
I've modified both the files but now it's failing, and I'm not entirely sure why, it mentions things that aren't in the generated src folder like BooleanValue..?

tiggerite:
So it was just nan itself being out of date.. I also updated node-pre-gyp for good measure. (There's an updated node-gyp as well but 4.0.0 is quite recent compared to the node-pre-gyp that was being used)

rcjsuen:
What a time to be alive...
Thank you for looking into this, @tiggerite!

tiggerite:
All should be good now, no more depreciations that I can see or find. Hopefully this will help get 0.25.0 out of alpha stage - it's already really stable as far as I can tell..

tiggerite:
@implausible is this all good now? It passes all the CI checks (cleanly too.. 😀)

implausible:
Oof, this seems to have introduced instability into the test suite. These tests are failing on master and on several unrelated PRs:

https://travis-ci.org/nodegit/nodegit/jobs/559068365

tiggerite:
That’s very strange. It seems to affect three Rebase and two Remote tests. I can’t see why the changes would be causing those to be flaky.

tiggerite:
Bit of a long shot but could it be because the enhanced tests still run on Node 8? Maybe try them on 10, the new Nan can’t be optimised (at the very least) for 8?

implausible:
Definitely worth considering. I've also re-read this entire PR and also can't determine what did it. I am going to double check that these tests hadn't been failing randomly before this PR as well. I don't recall seeing these tests fail in the past.

PR #1701, opened 2019-07-16
Exclude node v12 from package engines

chrisdothtml:


chrisdothtml:
Although, it looks like v12 support may have been added already in master, so this PR might be irrelevant now

implausible:
Node 12 support is coming in the next release as it's already been completed/merged.

PR #1717, opened 2019-08-30
Fix rebase using in-memory index

henkesn:
In the current master there is missing the `RebaseOptions#inmemory` flag. This is why rebasing is currently not possible at all on bare repos. This fix adds the missing field in supplements config file. Rebasing works fine with `Rebase.init`.

### DIFF
--- a/generate/input/libgit2-supplement.json
+++ b/generate/input/libgit2-supplement.json
@@ -1392,6 +1392,10 @@
             {
               "type": "void *",
               "name": "payload"
+            },
+            {
+              "type": "int",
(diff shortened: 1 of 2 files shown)


implausible:
Would you write a test case that utilizes the inmemory field?

henkesn:
Hm, the build seems broken.

implausible:
Yup. It looks like github actions aren't quite ready to handle pull_requests right now. https://github.com/actions/checkout/issues/23

henkesn:
Thank you! 👍

Issue #1359, opened 2017-08-24
DiffList is not returned


eliranmal:
it seems that both `Diff.treeToTree` and `Tree#diff` fail to return a `DiffList`, but instead return a `Diff` object.

the documentation [states][1] that they should, so i'm wondering - is `DiffList` not supported anymore? was it replaced by another type?

_by the way, that also means that the "diff-commits" example is broken, as [it iterates over such list][2]._

[1]: http://www.nodegit.org/api/tree/#diff
[2]: https://github.com/nodegit/nodegit/blob/master/examples/diff-commits.js#L22

eliranmal:
here's a code snippet, to put it in context:

    let repository;

    Git.Repository.open(gitRoot)
        .then(repo => {
            repository = repo;
            return repo;
        })
        .then((repo) => {
            return Promise.all([
                repo.getHeadCommit(),
                repo.getMasterCommit(),
            ])
        })
        .then(([head, master]) => {
            if (!master.id().equal(head.id())) {
                return Promise.all([
                    head.getTree(),
                    master.getTree(),
                ])
            }
            throw new Error ('no diff!');
        })
        .then(([headTree, masterTree]) => {
            
            // both of these will return a Diff, not a DiffList:
            
            // return Git.Diff.treeToTree(repository, headTree, masterTree);
            return headTree.diff(masterTree);
        })
        .then(list => {
            list.forEach((v, k) => { // --> TypeError: list.forEach is not a function
                console.log(`v: ${v}, k: ${k}`);
            });
        })

mariecl:
I am facing the same issue with very similar code.

rcjsuen:
I can run [`diff-commit.js`](https://github.com/nodegit/nodegit/blob/952bf8d244b2f9fa412c838f40e1a1d15bb0c9ea/examples/diff-commits.js) just fine so I don't think it's broken.

`Commit`'s `getDiff` function will return a `Diff[]` so there isn't a problem there.

eliranmal:
@rcjsuen - perhaps `Commit`'s API returns an array of `Diff`s, but that's not what this issue is about.

the issue here is other API entries (namely `Diff.treeToTree` and `Tree#diff`) that are either poorly documented or flat out fail to return their declared return type.

rcjsuen:
@eliranmal In your [original comment](https://github.com/nodegit/nodegit/issues/1359#issue-252552045), you mentioned:

[...]

PR #1733, opened 2019-10-06
Remove DiffList

igncp:
The `DiffList` is a legacy name for a type that doesn't currently exist. In cases where several diffs are returned, `Array<Diff>` is documented.

However, Tree#diff and Tree#diffWithOptions return a single `{Diff}`, which is incorrect in the docs:

- https://www.nodegit.org/api/tree/#diff
- https://www.nodegit.org/api/tree/#diffWithOptions

This PR removes `DiffList` from the repo to avoid confusion and updates to the correct type. Please see this previous issue for more context: https://github.com/nodegit/nodegit/issues/1359

Thanks for the library 👍

### DIFF
--- a/lib/tree.js
+++ b/lib/tree.js
@@ -34,7 +34,7 @@ Tree.prototype.builder = function() {
  * @async
  * @param {Tree} tree to diff against
  * @param {Function} callback
- * @return {DiffList}
+ * @return {Diff}
  */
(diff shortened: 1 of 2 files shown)


tbranyen:
Thanks, looks good!

implausible:
A wild Tim appeared!

Issue #1774, opened 2020-05-24
Make nodegit context-aware for compatibility with Electron 9 and beyond


Livven:
[Electron 9 changed](https://www.electronjs.org/blog/electron-9-0#breaking-changes) the default value of `app.allowRendererProcessReuse` to true, which breaks nodegit. Setting it to false manually fixes the issue, but this will only be possible [until Electron 11](https://github.com/electron/electron/blob/9bc5e9823821fc6fef4918f17d591408646b65e5/docs/breaking-changes.md#default-changed-loading-non-context-aware-native-modules-in-the-renderer-process-is-disabled-by-default).

Are there any plans to fix this? Seems like #1298 might be related. Also see https://github.com/electron/electron/issues/18397 for more information.

## System information
  - node version: v12.16.3
  - npm or yarn version: 6.14.4
  - OS/version/architecture: win32 10.0.18363 x64
  - Applicable nodegit version: 0.26.5

```
node -v
npm -v # (or yarn -v)
node -e "console.log(process.platform)"
node -e "console.log(require('os').release())"
node -e "console.log(console.log(process.arch))"
```

oyed:
This definitely needs to be looked in to, if not prioritized, for the future of `nodegit` in Electron environments given how fast Electron releases are rolling out. With RC for Electron 10 already out, it's only a matter of (short) time before 11 is the norm.

maxkorp:
Is there a preference for context-awareness over n-api compat?

implausible:
Axosoft is aware that we need to update this shortly. We'll be putting 2 engineers on this after our next release. Definitely prefer switching to n-api compat, that seems like a better long term. There's also a ton of utility in n-api that should simplify a lot of extra work we've had to do (thread safe callbacks, I'm looking at you).

Livven:
Awesome, good to hear that!

@maxkorp To be honest I have no idea what the difference and generally the context here is, I just used the terminology I saw in the warning messages :p

implausible:
[...]

PR #1788, opened 2020-08-13
Bump libgit2 to fork of v1.0.0

implausible:
Adds the following unmerged PRs from Libgit2:
- https://github.com/libgit2/libgit2/pull/5384
- https://github.com/libgit2/libgit2/pull/5602
- https://github.com/libgit2/libgit2/pull/5601
- https://github.com/libgit2/libgit2/pull/4205
- https://github.com/libgit2/libgit2/pull/5347

This should lay the groundwork for having correct async resources in the context aware version of NodeGit thanks to the custom tls implementation added to libgit2.

This PR is necessary to move https://github.com/nodegit/nodegit/issues/1774 forward more


implausible:
Well that test failure is consistently spooky.

implausible:
Of course it's not reproducible locally.

implausible:
Holy hell that is gnarly, but I don't think it's us, I think something is wrong with mocha on Node.

Here's what happens:
1. RunLoopCallbacks is scheduled on the event loop
2. credentials_async is called
3. ForwardIfPromise is called, it detects that it is a promise, mounts the callbacks on `then`, and exits with true
4. credentials_async exits
5. RunLoopCallbacks exits
6. The promise chain never starts.
7. Mocha cancels the test
8. The promise chain starts
9. The promise chain finishes
10. credentials_promiseCompleted is called and finishes

The only way I think this could happen is if mocha had shceduled something that is scheduled to run _before_ the promise chain built in the credentials callback, and then it blocks the javascript thread until mocha cancels it. (mocha must be in control here, because mocha is able to cancel the test).

Fucking wild.

I've temporarily disabled that portion of the test... I am unable to place how this would be NodeGit's fault.

implausible:
Oof... Ok, I think it is our fault after all. I think it's related to:
https://github.com/nodejs/node/issues/5691

I think we need to fix async resources to solve this issue, which means that this bump to 1.0.0 can't just be a bump to 1.0.0, it needs to also fix https://github.com/nodegit/nodegit/issues/1774. I was hoping to bring in the patches we need in libgit2 into NodeGit before fixing async resources, but with this issue being so fucking consistent, I think there's no other way.

Also, I noticed that this behavior can occur outside of node 14 x64 on windows, I've seen this exact behavior tank the rebase commit signing callback tests on Node 10. I think this might be a case of undefined behavior and a game of wacky wavy inflatable timeouts.

All I know is the second that timeout callback occurs, it allows the frozen promise chain to start running.

implausible:
Ok after a pretty intense refactor of the ThreadPool to support AsyncResource passing, I discovered that it's not at fault. I still needed to do the refactor to get AsyncResource passing as part of the Context Awareness refactor, so that's fine. What it all boiled down to was that we need to initialize a `node::CallbackScope` when we enter the libuv event loop. Without running this, all of the things that node does in the background after coming back from a thread aren't scheduled to run. In a bizarre circumstance, this _always triggered_ on Windows 10 x64 Node 14.

This issue roughly explains the problem https://github.com/nodejs/node/issues/5691. I don't know why we don't run into this more often, but it's definitely not a problem in nodegit anymore. Wild.

implausible:
I'm going to take that fix and then split out the thread pool fix into another PR for Context-Awareness

PR #1795, opened 2020-09-11
Refactor for context-awareness

### LINKED ISSUE #1774: Make nodegit context-aware for compatibility with Electron 9 and beyond

ianhattendorf:
TLDR:
- `worker_thread` support!
- Electron support when `app.allowRendererProcessReuse=true`

Fixes #1774


generate/templates/manual/src/thread_pool.cc
```diff
@@ -672,12 +674,10 @@ namespace nodegit {
       jsThreadCallbackQueue.pop();
     }
 
-    // NOTE We are deliberately leaking this pointer because `async` cleanup in
-    // node has not completely landed yet. Trying to cleanup this pointer
-    // is probably not worth the fight as it's very little memory lost per context
-    // When all LTS versions of node and Electron support async cleanup, we should
-    // be heading back to cleanup this
-    uv_close((uv_handle_t *)jsThreadCallbackAsync, nullptr);
+    uv_close(reinterpret_cast<uv_handle_t *>(jsThreadCallbackAsync), [](uv_handle_t *handle) {
```
implausible, line 677: We're leaking the handle still

generate/templates/manual/src/thread_pool.cc
```diff
@@ -672,12 +674,10 @@ namespace nodegit {
...
+      auto threadPoolImpl = static_cast<ThreadPoolImpl *>(handle->data);
+      delete threadPoolImpl->currentContext;
```
implausible, line 679: I'm not really a fan of the threadpool deleting the context like this. Could we add a completion callback to the shutdown routine instead?

implausible:
hoooweee...

PR #1797, opened 2020-09-22
Update longpath enums to match libgit2

ianhattendorf:


ianhattendorf:
I want to add a test for this, hold off on reviewing for now.

test/tests/clone.js
```diff
@@ -54,6 +81,29 @@ describe("Clone", function() {
...
+    fse.ensureDirSync(longClonePath);
+
+    if (process.platform === "windows") {
+      NodeGit.Libgit2.opts(NodeGit.Libgit2.OPT.SET_WINDOWS_LONGPATHS, 1);
+    }
+
+    return Clone(url, longClonePath, opts).then(function(repo) {
+      assert.ok(repo instanceof Repository);
+      test.repository = repo;
+    });
```
implausible, line 104: This is not what the long paths option is about. The option you're looking to test is after a repository is _created_, you can add items to the index even though they have a long path.

ianhattendorf: This should be a valid test for longpaths. All file operations should work for long paths, including cloning a normal repo into a long path (or e.g. cloning a repo into a path with length 240 chars with a name putting it over 260 chars).

Tests were failing because I used `windows` instead of `win32` for `process.platform`.

implausible: Well shoot. Thanks for correcting me!

implausible (approved):

PR #1801, opened 2020-10-05
Remove block for sideband_progress in remote_callbacks

adarsh9pai:
Simple PR. Just removing block for sideband_progress in remote_callbacks from generate/input/descriptor.json.


implausible:
I see this, but we're kind of on a slight hiatus while we wait for Node to backport async cleanup routines. This will make it in shortly after that PR merges and we build a release.

implausible (approved):

Issue #1800, opened 2020-10-01
move the examples to async/await syntax


julien-c:
Any interest for this?

I can work on a PR if there's interest from the maintainers.

Would make things way cleaner, especially in cases where you need to move variables around from one promise to others in the chain.

implausible:
That would be ideal. Using an up-to-date ecma standard for our examples would be awesome.

PR #1802, opened 2020-10-09
First stab at #1800 (async/await in examples)

julien-c:
The only non-syntactic changes are:
- used built-in `fs` instead of `fse` as the methods are there already
- fixed a few deprecation notices

Feedback welcome

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -2,7 +2,7 @@
   "boss": true,
   "curly": true,
   "eqnull": true,
-  "esnext": true,
+  "esversion": 9,
   "evil": true,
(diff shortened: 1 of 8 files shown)


julien-c:
Anyone? :) @implausible @ianhattendorf 

Also explicitly referencing https://github.com/nodegit/nodegit/issues/1800

implausible:
Yes, we see this. I can't promise I will get to it immediately, but will try to land it before the next release. Thanks for your effort!

implausible (approved):
Thanks!

PR #1822, opened 2021-03-04
Don't strdup nullptr from git_tag_message 

ianhattendorf:
git_tag_message can return nullptr. From docs: "message of the tag or NULL when unspecified"

e.g. https://gitlab.freedesktop.org/mesa/mesa/-/tags/R300_DRIVER_0


implausible (approved):

implausible:
Need to open up a sidecar branch to release until we get electron 11 patched.

implausible (approved):
Nice find.

PR #1823, opened 2021-03-05
returns_info: fix ownedByIndices

wmertens:
I'm preparing a PR for using eslint and prettier, and eslint pointed out this unused variable.

It looks like this was a mistake?


wmertens:
cc @implausible, it seems like you added the code

implausible:
@wmertens Can you make this branch based on / targeting `release/0.28.0`. We're doing an intermediate release outside of `master` branch as we wait for https://github.com/electron/electron/issues/28043 to get solved.

implausible (approved):

PR #1824, opened 2021-03-06
Include libgit2 winhttp proxy fix

ianhattendorf:
See https://github.com/libgit2/libgit2/pull/5814


ianhattendorf:
Looking into failure, not sure why it's only failing on node 14. Maybe a race condition.

implausible:
I wonder if we can salvage master so that we can release from it if we just remove the broken context-awareness behavior.

PR #1826, opened 2021-03-09
Check if module is compiled under context aware node version

ianhattendorf:
If not, leak cleanup pointer like before.

Once the electron patch is in and released we can revert this.

https://github.com/electron/electron/issues/28043


implausible:
This is looking good. I am going to do some test compiles for it

PR #1835, opened 2021-04-14
Define optional arguments for Patch.fromBlobs()

weedz:
According to the libgit2 documentation (https://libgit2.org/libgit2/#HEAD/group/patch/git_patch_from_blobs) all arguments are allowed to be NULL.

### DIFF
--- a/generate/input/descriptor.json
+++ b/generate/input/descriptor.json
@@ -2743,6 +2743,29 @@
         "git_patch_free": {
           "ignore": true
         },
+        "git_patch_from_blobs": {
+          "isAsync": true,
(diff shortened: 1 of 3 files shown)


implausible (changes requested):
Would you kindly add a test case for calling `NodeGit.Patch.fromBlobs` with all optionals optional?

weedz:
> Would you kindly add a test case for calling `NodeGit.Patch.fromBlobs` with all optionals null?

Sure. Don't really know what the expected behavior is but seems to generate an "empty" patch object with size 0.

PR #1837, opened 2021-04-22
Eliminate need for normalize options

implausible:
This should eliminate like half of the crash issues we get asked on GitHub. Basic gist is that we've modified the way we read arguments on async functions such that in C++ we read _any_ js object and unpack the fields from the js object into a brand new options object at call time. After we build that object, we perform the async work, and when the async work is completed, we free the objects.

## Basic flow:
1. Async function is called
2. We recognize an argument as a ConfigurableClass
3. We unpack the Configurable class from the argument and create a brand new options object
4. We attach a shared_ptr to the ConfigurableClass on AsyncWorker
5. We perform the async work
6. When the async work is complete, we tear down the ConfigurableClasses that were built for the async function call

### FilterRegistry
1. When we register a filter in FilterRegistry, we save the ConfigurableClass on the current nodegit::Context.
2. When we unregister the filter, we remove the handle for the filter from the nodegit::Context

### Filter::Load
1. We fetch the list of filters necessary to build the filter list
2. We preserve the filters that the filter list uses on the FilterList, such that they do not go out of scope until FilterList does. We fetch these from the FilterRegistry on the nodegit::Context

### Rebase.open / Rebase.init / Remote.prototype.connect
We preserve the options objects _on_ the "this" object, since these objects need to live as long as the Rebase or the Remote after calling one of these functions


lib/deprecated/structs/ApplyOptions.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function() {
+  this.flags = 0;
```
AlexaXs, line 2: Only field 'flags' needed here? why not the rest of the fields of the structure in https://libgit2.org/libgit2/#HEAD/type/git_apply_options

AlexaXs (commented):
For what I've seen/understood, it looks good to me

lib/deprecated/structs/ApplyOptions.js
implausible: Those fields don't really benefit from having a default value. It's important that fields that are numbers are built out with a default value so that users who are depending on those items being numbers continue to work correctly.

The best example of this is somewhere in our codebase we do: `diffOptions.flags |= NodeGit.Diff.IGNORE_WHITESPACE`. But this would not work if flags is not already a number, as `undefined | 5` produces `NaN`.

If we compare that to any of the other fields that are not numbers, the default values for these are `null`. In JS we're basically just expressing `null` as `undefined` by the field not being on the object.

PR #1846, opened 2021-06-09
Statistics with same output as "git-sizer -j"

AlexaXs:
# Statistics

It returns same statistics as executing 'git-sizer -j' (https://github.com/github/git-sizer).

An example of the output of a test repository:
``` yaml
{
  "repositorySize": {
    "commits": {
      "count": 992,
      "size": 265544
    },
    "trees": {
      "count": 2416,
      "size": 1188325,
      "entries": 32571
    },
    "blobs": {
      "count": 4149,
      "size": 48489622
    },
    "annotatedTags": {
      "count": 1
    },
    "references": {
      "count": 10
    }
  },
  "biggestObjects": {
    "commits": {
      "maxSize": 956,
      "maxParents": 2
    },
    "trees": {
      "maxEntries": 93
    },
    "blobs": {
      "maxSize": 1077756
    }
  },
  "historyStructure": {
    "maxDepth": 931,
    "maxTagDepth": 1
  },
  "biggestCheckouts": {
    "numDirectories": 128,
    "maxPathDepth": 10,
    "maxPathLength": 107,
    "numFiles": 514,
    "totalFileSize": 5160886,
    "numSymlinks": 2,
    "numSubmodules": 4
  }
}
```
## repositorySize

It counts all the repository's unique objects (and references).

## biggestObjects

The maximum objects found. For instance the value 93 in 'biggestObjects.trees.maxEntries' means that from all the unique trees in the repository, the tree with the maximum number of entries has 93 entries.

### maxDepth

This analyzes the commit graph of all the history of the repository, and it's the path with the maximum number of commits from an initial commit to one of the latest commits.

It considers that a graph can have multiple initial commits.

### maxTagDepth

This is the maximum number of tags chained until a non-tag object is reached.

## biggestCheckouts

This is the result of comparing all the commits in the history of the repository, and returning the maximum values found. The result does not have to belong to a unique commit, for instance a commit C1 could have the maximum number of directories (128 in the example), while other commit C2 has the maximum sum of files sizes it contains (5160886 in the example).

The numbers shown are the result of considering the entries in the commit's tree and also all its subtrees.

# Notes about the result

The output return avoids considering dangling (unreachable) objects, as well as empty trees (like the one git introduces internally "4b825dc642cb6eb9a060e54bf8d69288fbee4904")

### DIFF
--- a/generate/templates/manual/include/worker_pool.h
+++ b/generate/templates/manual/include/worker_pool.h
@@ -1,0 +1,161 @@
+#ifndef WORK_POOL_H
+#define WORK_POOL_H
+
+#include <type_traits>
+#include <functional>
+#include <memory>
+#include <queue>
+#include <vector>
+#include <condition_variable>
+#include <mutex>
+#include <thread>
+
+/**
+ * \class WorkItem
+ * Abstract class for work items in the WorkerPool.
+ */
+class WorkItem
+{
+public:
+  WorkItem() = default;
+  virtual ~WorkItem() = default;
+  WorkItem(const WorkItem &other) = default;
+  WorkItem(WorkItem &&other) = default;
+  WorkItem& operator=(const WorkItem &other) = default;
+  WorkItem& operator=(WorkItem &&other) = default;
+};
+
+/**
+ * \class Worker
+ * Interface for Workers in the WorkerPool.
+ */
+class IWorker
+{
+public:
+  IWorker() = default;
+  virtual ~IWorker() = default;
+  IWorker(const IWorker &other) = delete;
+  IWorker(IWorker &&other) = delete;
+  IWorker& operator=(const IWorker &other) = delete;
+  IWorker& operator=(IWorker &&other) = delete;
+
+  virtual bool Initialize() = 0;
+  virtual bool Execute(std::unique_ptr<WorkItem> &&work) = 0;
+};
+
+/**
+ * \class WorkerPool
+ * To leverage this class, a Worker must inherit from IWorker.
+ * WorkItem is an abstract class from which to inherit too.
+ */
+template<class Worker, class WorkItem>
+class WorkerPool {
+public:
+  WorkerPool();
+  ~WorkerPool() = default;
+  WorkerPool(const WorkerPool &other) = delete;
+  WorkerPool(WorkerPool &&other) = delete;
+  WorkerPool& operator=(const WorkerPool &other) = delete;
+  WorkerPool& operator=(WorkerPool &&other) = delete;
+
+  bool Init(std::vector< std::shared_ptr<Worker> > workers);
+  bool InsertWork(std::unique_ptr<WorkItem> &&work);
+  void Shutdown();
+
+private:
+  void DoWork(std::shared_ptr<Worker> worker);
+  
+  std::mutex m_workQueueMutex {};
+  std::condition_variable m_workQueueCondition {};
+  std::queue< std::unique_ptr<WorkItem> > m_workQueue {};
+  std::vector<std::unique_ptr<std::thread>> m_threads {};
+  bool m_stop {true}; // initially the workpool has no worker threads
+};
+
+
+template<class Worker, class WorkItem>
+WorkerPool<Worker,WorkItem>::WorkerPool() {
+  static_assert(std::is_base_of<IWorker, Worker>::value, "Worker must inherit from IWorker");
+}
+
+// returns true it it wasn't initialized previously; false otherwise
+template<class Worker, class WorkItem>
+bool WorkerPool<Worker,WorkItem>::Init(std::vector< std::shared_ptr<Worker> > workers)
+{
+  {
+    std::lock_guard<std::mutex> lock(m_workQueueMutex);
+    if (!m_stop)
+      return false;
+    m_stop = false;
+  }
+  
+  std::for_each (workers.begin(), workers.end(), [this](std::shared_ptr<Worker> worker) {
+    m_threads.emplace_back(std::make_unique<std::thread>(std::bind(&WorkerPool::DoWork, this, worker)));
+  });
+  
+  return true;
+}
+
+// returns true if successfully inserted work; false otherwise
+template<class Worker, class WorkItem>
+bool WorkerPool<Worker,WorkItem>::InsertWork(std::unique_ptr<WorkItem> &&work)
+{
+  {
+    std::lock_guard<std::mutex> lock(m_workQueueMutex);
+    if (m_stop)
+      return false;
+    m_workQueue.emplace(std::move(work));
+  }
+  m_workQueueCondition.notify_one();
+  
+  return true;
+}
+
+template<class Worker, class WorkItem>
+void WorkerPool<Worker,WorkItem>::Shutdown()
+{
+  {
+    std::lock_guard<std::mutex> lock(m_workQueueMutex);
+    if (m_stop)
+      return;
+    m_stop = true;
+  }
+  m_workQueueCondition.notify_all();
+  
+  std::for_each (m_threads.begin(), m_threads.end(), [](std::unique_ptr<std::thread> &wt) {
+    if (wt->joinable()) {
+      wt->join();
+    }
+  });
+}
+
+template<class Worker, class WorkItem>
+void WorkerPool<Worker,WorkItem>::DoWork(std::shared_ptr<Worker> worker)
+{
+  if (!worker->Initialize()) {
+    return; // signal main that we've exited early
+  }
+  
+  while (true) {
+    std::unique_ptr<WorkItem> work {};
+    {
+      std::unique_lock<std::mutex> lock(m_workQueueMutex);
+      m_workQueueCondition.wait(lock, [this] {
+        return this->m_stop || !this->m_workQueue.empty();
+      });
+      
+      if (m_stop && m_workQueue.empty())
+        return;
+      
+      work = std::move(m_workQueue.front());
+      m_workQueue.pop();
+    }
+    
+    if (!worker->Execute(std::move(work))) {
+      return; // signal main that we've exited early
+    }
+  }
+}
+
+#endif  // WORK_POOL_H
+
--- a/generate/templates/manual/include/worker_pool.h
+++ b/generate/templates/manual/include/worker_pool.h
@@ -9,6 +9,7 @@
 #include <condition_variable>
 #include <mutex>
 #include <thread>
+#include <atomic>
 
 /**
  * \class WorkItem
@@ -43,6 +44,14 @@ public:
   virtual bool Execute(std::unique_ptr<WorkItem> &&work) = 0;
 };
 
+/* Enumeration describing the Worker Pool Status:
+* - kOk: everything ok.
+* - kInitializeFailed: a worker thread failed when calling Initialize().
+* - kExecuteFailed: a worker thread failed when calling Execute().
+* - kShutdownEarly: InsertWork() was called but the worker pool was stopped.
+*/
+enum class WPStatus {kOk, kInitializeFailed, kExecuteFailed, kShutdownEarly};
+
 /**
  * \class WorkerPool
  * To leverage this class, a Worker must inherit from IWorker.
@@ -61,25 +70,17 @@ public:
   void Init(std::vector< std::shared_ptr<Worker> > workers);
   void InsertWork(std::unique_ptr<WorkItem> &&work);
   void Shutdown();
-  bool Status() { return m_status.workersInitialize && m_status.workersExecute && !m_status.shutdownEarly; }
-  bool StatusWorkersInitialize() { return m_status.workersInitialize; }
-  bool StatusWorkersExecute() { return m_status.workersExecute; }
-  bool StatusShutdownEarly() { return !m_status.shutdownEarly; }
+  WPStatus Status() { return m_atomicWPStatus; }
 
 private:
   void DoWork(std::shared_ptr<Worker> worker);
   
-  std::mutex m_workQueueMutex {};
-  std::condition_variable m_workQueueCondition {};
+  std::mutex m_mutex {};            // locks m_workQueue and m_stop
+  std::condition_variable m_condition {};
   std::queue< std::unique_ptr<WorkItem> > m_workQueue {};
-  std::vector<std::unique_ptr<std::thread>> m_threads {};
   bool m_stop {true};               // initially the workpool has no worker threads
-  struct {
-    bool workersInitialize {true};
-    bool workersExecute {true};
-    bool shutdownEarly {false};   // for instance if the workerPool was ShutDown, and InsertWork called after that
-  } m_status;
-  std::mutex m_statusMutex {};
+  std::vector<std::unique_ptr<std::thread>> m_threads {};
+  std::atomic<WPStatus> m_atomicWPStatus {WPStatus::kOk};
 };
 
 
@@ -93,7 +94,7 @@ template<class Worker, class WorkItem>
 void WorkerPool<Worker,WorkItem>::Init(std::vector< std::shared_ptr<Worker> > workers)
 {
   {
-    std::lock_guard<std::mutex> lock(m_workQueueMutex);
+    std::lock_guard<std::mutex> lock(m_mutex);
     if (!m_stop)
       return;
     m_stop = false;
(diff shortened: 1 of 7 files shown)


implausible (changes requested):

generate/templates/manual/include/worker_pool.h
```diff
@@ -0,0 +1,161 @@
...
+      wt->join();
+    }
+  });
+}
+
+template<class Worker, class WorkItem>
+void WorkerPool<Worker,WorkItem>::DoWork(std::shared_ptr<Worker> worker)
+{
+  if (!worker->Initialize()) {
+    return; // signal main that we've exited early
```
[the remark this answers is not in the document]
AlexaXs: Effectively, that part wasn't complete, workerPool was lacking a status.
Commited changes: now the status is updated on Init(), when each worker calls Initialize(), and when each worker calls Execute(). The status can be checked by the workerPool user.

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+
+/**
+ * CommitsGraph::AddNode
+ * 
+ * \param oidStr oid of the commit object to add.
+ * \param parents oids of the commit's parents.
+ * \param numParents Number of parents of the commit object.
+ */
+void CommitsGraph::AddNode(const std::string &oidStr, const std::vector<std::string> &parents,
+  uint32_t numParents)
```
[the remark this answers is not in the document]
AlexaXs: You are right, and probably pushing an extra parameter to the stack in the call to AddNode is more expensive than calculating numParents again from the vector size (end() - begin())

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+  else {
+    if (git_repository_open(&m_repo, m_repoPath.c_str()) != GIT_OK) {
+      return false;
+    }
+    
+    if (git_repository_odb(&m_odb, m_repo) != GIT_OK) {
+      return false;
+    }
+  }
+  return true;
```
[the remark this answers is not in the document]
AlexaXs: Absolutely, shorter and still readable, thanks!

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+  pruneUnreachableCommits();
+  pruneUnreachableTrees();
+  pruneUnreachableBlobs();
+}
+
+/**
+ * RepoAnalysis::pruneUnreachableTags
+ * Prune tags and their chained tags if they become unreachable.
+ * Also decreases reachability of targets.
+  */
```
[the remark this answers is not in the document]
AlexaXs: If you follow these steps, you'll end up with an unreachable tag:

1. Add an annotated tag 'tagTest' to a commit
   - A file is also added (".git/refs/tags/tagTest") containing the sha-1 of the annotated tag object just created: bc7544879cdccc964e4e6a1489ec54c030447f81
   - Executing command "git cat-file -p bc7544879cdccc964e4e6a1489ec54c030447f81" shows the commit target of the tag:
    object 113f9e324b531c12aca4b26fd0d23cd41eb93899
    type commit

2. Delete the annotated tag created 'tagTest':
  - File in ".git/refs/tags/tagTest" is deleted. Now the tag is not reachable.
  - Executing command "git cat-file -p bc7544879cdccc964e4e6a1489ec54c030447f81" still shows the same information as in step 1, meaning that the annotated commit object bc7544879cdccc964e4e6a1489ec54c030447f81 still exists inside the objects database ".git/objects".
  - We can also check it executing "git prune -n" (https://git-scm.com/docs/git-prune) which will show the unreachable objects. It returns: "bc7544879cdccc964e4e6a1489ec54c030447f81 tag".
  - bc7544879cdccc964e4e6a1489ec54c030447f81 has become an unreachable or dangling tag.

generate/templates/manual/include/worker_pool.h
```diff
@@ -58,9 +58,13 @@ class WorkerPool {
...
 
-  bool Init(std::vector< std::shared_ptr<Worker> > workers);
-  bool InsertWork(std::unique_ptr<WorkItem> &&work);
+  void Init(std::vector< std::shared_ptr<Worker> > workers);
+  void InsertWork(std::unique_ptr<WorkItem> &&work);
   void Shutdown();
+  bool Status() { return m_status.workersInitialize && m_status.workersExecute && !m_status.shutdownEarly; }
+  bool StatusWorkersInitialize() { return m_status.workersInitialize; }
+  bool StatusWorkersExecute() { return m_status.workersExecute; }
+  bool StatusShutdownEarly() { return !m_status.shutdownEarly; }
```
implausible, line 67: Mutex is required to read, not just write. Compiler optimizations can prevent these shared values from ever being read. See https://devblogs.microsoft.com/oldnewthing/20180924-00/?p=99805.

I would instead leverage `std::atomic<bool>` instead of leveraging any mutexes here at all.

implausible (changes requested):

generate/templates/manual/include/worker_pool.h
AlexaXs: Good catch! thank you!

In the fix I've reduced the status to just one atomic enumeration variable

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+    {
+      OdbObjectsData::iterCommitInfo itCommitInfo =
+      m_odbObjectsData->commits.info.find(tagInfo->oidTarget);
+    
+      if (itCommitInfo != m_odbObjectsData->commits.info.end()) {
+        { // lock
+          std::lock_guard<std::mutex> lock(m_odbObjectsData->infoMutex.commits);
+          ++itCommitInfo->second.reachability;
+        }
+      }
```
ianhattendorf, line 703: There's a data race here: `m_odbObjectsData->commits.info` is read here without a lock, and possibly written to at the same time via `WorkerStoreOdbData::Execute`. Standard collections may be either read from multiple threads concurrently or written to by a single thread, but not both.

ianhattendorf: Same with the tree and blob sections below.

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+
+  // add 1 to its tree's reachability
+  OdbObjectsData::iterTreeInfo itCommitTreeInfo =
+    m_odbObjectsData->trees.info.find(commitInfo->oidTree);
+  
+  if (itCommitTreeInfo != m_odbObjectsData->trees.info.end()) {
+    { // lock
+      std::lock_guard<std::mutex> lock(m_odbObjectsData->infoMutex.trees);
+      ++itCommitTreeInfo->second.reachability;
+    }
```
ianhattendorf, line 781: Data race 2x

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+
+  // set entry trees' reachability
+  for (auto &treeNameLen : treeInfo->entryTreesNameLen) {
+    OdbObjectsData::iterTreeInfo itTreeInfo = m_odbObjectsData->trees.info.find(treeNameLen.first);
+    
+    if (itTreeInfo != m_odbObjectsData->trees.info.end()) {
+      { // lock
+        std::lock_guard<std::mutex> lock(m_odbObjectsData->infoMutex.trees);
+        ++itTreeInfo->second.reachability;
+      }
```
ianhattendorf, line 814: Data race 2x

ianhattendorf (changes requested):
Still working my way through but figured I'd leave some partial feedback, overall this is looking really good :ok_hand:

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+  }
+
+  while (!children.empty()) {
+    ++maxDepth;
+    parents = std::move(children);
+
+    // add unique children of next level, and only if from the last parent
+    for (CommitsGraphNode *parent : parents) {
+      for (CommitsGraphNode *child : parent->children) {
+        if (--child->parentsLeft == 0) {
```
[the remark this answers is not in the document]
AlexaXs: I'll add better comments in the code in this part, since the algorithm needs explaining.

Inside `CalculateMaxDepth()` we are only modifying the `parentsLeft` of each node. At the begining of each iteration, the nodes from `children` are being moved into `parents`, so all the nodes in the graph starting from `m_roots` are being modified intentionally.

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+  WorkerStoreOdbData(const WorkerStoreOdbData &other) = delete;
+  WorkerStoreOdbData(WorkerStoreOdbData &&other) = delete;
+  WorkerStoreOdbData& operator=(const WorkerStoreOdbData &other) = delete;
+  WorkerStoreOdbData& operator=(WorkerStoreOdbData &&other) = delete;
+
+  bool Initialize();
+  bool Execute(std::unique_ptr<WorkItem> &&work);
+
+private:
+  OdbObjectsData::TreeInfoAndStats thisTreeInfoAndStats(git_tree *tree, size_t size, size_t numEntries);
```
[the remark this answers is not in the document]
AlexaXs: Yes indeed! 👍

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1889 @@
...
+ * \return this tree's data and partial statistics.
+  */
+OdbObjectsData::TreeInfoAndStats WorkerStoreOdbData::thisTreeInfoAndStats(git_tree *tree, size_t size,
+  size_t numEntries)
+{
+  const git_tree_entry *te {nullptr};
+  git_object_t te_type {GIT_OBJECT_INVALID};
+  const char *teName {nullptr};
+  size_t teNameLen {0};
+  const git_oid *te_oid {nullptr};
```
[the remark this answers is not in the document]
AlexaXs: You are right, thanks!

generate/templates/manual/repository/statistics.cc
AlexaXs: I've also added better comments to try to clarify the code.

To obtain the final result, the whole process is run in different stages:

>  // stage 1 methods: store data from repository (with threads)
>  int storeObjectsInfo();
>  int storeAndCountRefs();
>  // stage 2 methods: count reachability of each object (with threads)
>  // NOTE: we need this stage, since so far libgit2 doesn't provide unreachable objects
>  bool setObjectsReachability();
>  void setReachabilityFromRefs();
>  void setUnreachables();
>  // stage 3 methods: prune unreachable oids
>  void pruneUnreachables();
>  void pruneUnreachableTags();
>  void pruneUnreachableCommits();
>  void pruneUnreachableTrees();
>  void pruneUnreachableBlobs();
>  // stage 4 methods: repositorySize and biggestObjects
>  void statsCountAndMax();
>  // stage 5 methods: historyStructure and biggestCheckouts
>  bool statsHistoryAndBiggestCheckouts();
>  bool calculateBiggestCheckouts();
>  OdbObjectsData::iterTreeInfo calculateTreeStatistics(const std::string &oidTree);
>  bool calculateMaxTagDepth();
>  OdbObjectsData::iterTagInfo calculateTagDepth(const std::string &oidTag);

If a stage leverages threads via a worker pool, the worker pool is created and we wait until all the threads are done to continue with the next stage.

Stage 1 leveraged threads that run `WorkerStoreOdbData::Execute`, storing information in `m_odbObjectsData->commits.info`, `m_odbObjectsData->trees.info`, etc.

In this case at this point of the code we are in stage 2, where `WorkerReachCounter::Execute` is called instead of `WorkerStoreOdbData::Execute`, and threads from stage 1 have all stopped, hence there won't be a data race with the containers written during stage 1.

Thanks! since this needed a better explanation in the code 👍

generate/templates/manual/repository/statistics.cc
AlexaXs: same as above

generate/templates/manual/repository/statistics.cc
AlexaXs: same as above

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1912 @@
...
+    // erase unreachable commits
+    for (OdbObjectsData::iterUnreachable itCommitUnrch = m_odbObjectsData.commits.unreachables.begin();
+      itCommitUnrch != m_odbObjectsData.commits.unreachables.end(); ++itCommitUnrch)
+    {
+      OdbObjectsData::iterCommitInfo itCommitInfo = m_odbObjectsData.commits.info.find(*itCommitUnrch);
+    
+      if (itCommitInfo != m_odbObjectsData.commits.info.end())
+      {
+        // decrease commit's parents reachability and add them as newUnreachable
+        size_t numParents = itCommitInfo->second.parents.size();
```
ianhattendorf, line 1351: const

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1912 @@
...
+ * Statistics for repositorySize (count objects) and biggestObjects (get maximum of them).
+ * Also builds the commits graph.
+ * NOTE: better results achieved not leveraging threads.
+ */
+void RepoAnalysis::statsCountAndMax()
+{
+  // commits
+  for (auto &info : m_odbObjectsData.commits.info) {
+    OdbObjectsData::CommitInfo &commitInfo = info.second;
+    size_t size = commitInfo.size;
```
ianhattendorf, line 1447: I would search `size_t size` and make all non-member non-param instances const.

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1912 @@
...
+      m_odbObjectsData.commits.maxParents, commitInfo.parents.size());
+
+    // build commit's graph
+    m_odbObjectsData.commits.graph.AddNode(info.first, commitInfo.parents);
+  }
+  // trees
+  for (auto &info : m_odbObjectsData.trees.info) {
+    OdbObjectsData::TreeInfoAndStats &treeInfo = info.second;
+    size_t size = treeInfo.size;
+    size_t numEntries = treeInfo.numEntries;
```
ianhattendorf, line 1461: Same with  `size_t numEntries`

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1912 @@
...
+  for (auto &oidBlob : treeInfoAndStats.entryBlobs)
+  {
+    if ((itBlobInfo = m_odbObjectsData.blobs.info.find(oidBlob)) == m_odbObjectsData.blobs.info.end()) {
+      return m_odbObjectsData.trees.info.end(); // to let the caller know that something went wrong
+    }
+
+    treeInfoAndStats.stats.totalFileSize += itBlobInfo->second.size;
+  }
+
+  // recursively into subtrees
```
ianhattendorf, line 1580: Is there any chance of stack overflow here?

ianhattendorf (changes requested):

generate/templates/manual/repository/statistics.cc
ianhattendorf: I see, yes that helps clarify :+1:

generate/templates/manual/repository/statistics.cc
```diff
@@ -0,0 +1,1912 @@
...
+
+  if ((errorCode = git_odb_foreach(odb, forEachOdbCb, &workerPool)) != GIT_OK) {
+    workerPool.Shutdown();
+    git_odb_free(odb);
+    return errorCode;
+  }
+
+  // main thread will work on the refs while waiting for the threads to finish
+  if ((errorCode = storeAndCountRefs() != GIT_OK)) {
+    return errorCode;
```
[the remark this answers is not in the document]
AlexaXs: Yes! thank you for caching this!

generate/templates/manual/repository/statistics.cc
AlexaXs: In this case it can be taken out to the method scope and reused for all the loops

generate/templates/manual/repository/statistics.cc
AlexaXs: The maximum number of recursive calls depend directly on the maximum path depth of the repository. For instance, the linux repository have a maximum path depth of 13, so I'd say the chances of stack overflow are pretty low here, but it was very good to have it double checked 👍 Thanks!

ianhattendorf (approved):

PR #1859, opened 2021-09-15
Fix a reference error when compiling with VC2019

yjoer:
I encountered an error when building from the source (in the master branch) in Windows 10 using VC2019 v16.11.3.

```
path\to\nodegit\include\cleanup_handle.h(16,19): error C2039: 'string': is not a member of 'std'
```

Here's my fix by including the `<string>` header, and the build turns successful after re-running the `npm run rebuild` command.


yjoer:
It is now basing on the master branch.

PR #1866, opened 2021-10-14
node-pre-gyp -> @mapbox/node-pre-gyp

samuela:
The `node-pre-gyp` has been deprecated, moving to `@mapbox/node-pre-gyp` instead. See [here](https://www.npmjs.com/package/node-pre-gyp) and [here](https://www.npmjs.com/package/@mapbox/node-pre-gyp) for more info.

Built successfully and all tests pass.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,9 +1,7857 @@
 {
   "name": "nodegit",
   "version": "0.28.0-alpha.9",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "version": "0.28.0-alpha.9",
+      "hasInstallScript": true,
+      "license": "MIT",
+      "dependencies": {
+        "@mapbox/node-pre-gyp": "^1.0.5",
+        "fs-extra": "^7.0.0",
+        "got": "^10.7.0",
+        "json5": "^2.1.0",
+        "lodash": "^4.17.14",
+        "nan": "^2.14.1",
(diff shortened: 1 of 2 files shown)


jmurzy:
How can we get this into the next v28-alpha release?

jfirebaugh:
There was [a breaking change in 1.0.0](https://github.com/mapbox/node-pre-gyp/blob/40a83f20a762668c41153eaeef8574373f70fe30/CHANGELOG.md#100) that affects the GitHub action
> no longer supporting `node_pre_gyp_accessKeyId` & `node_pre_gyp_secretAccessKey`, use `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` instead to authenticate against s3 for `info`, `publish`, and `unpublish` commands.

This code will need to be updated:

https://github.com/nodegit/nodegit/blob/bae468cf3da6ae5ef6d3f89d11bf89f83e8db94a/.github/workflows/tests.yml#L69-L71

And similarly for macOS and Windows.

ianhattendorf:
Is `node-pre-gyp` causing issues for anyone right now? I understand it's deprecated, but we have quite a bit on our backlog so if it's not broken we'll prioritize other tasks over updating packages that are still functioning.

samuela:
This was a while ago so I don't remember what is/was breaking, but node-pre-gyp has been deprecated for over a year now. Prioritization is ofc up to maintainers tho.

ianhattendorf:
This was handled alongside #1897, but thanks for bringing this to our attention!

PR #1883, opened 2021-12-14
UNSAFE Temporary workaround for LFS checkout performance regression

AlexaXs:
After a Threadpool rewrite for context-awareness, LFS checkouts started suffering long delays. We found the reason is that libgit2 callbacks that were being run in parallel started being run in sequence. But this is actually the safest way to deal with the callbacks, preventing deadlocks while being executed, and keeping the non-callback work thread-safe.

The changes here will just be an `UNSAFE` temporary workaround until LFS is properly integrated into NodeGit. The changes will be limited to the processing of callbacks from Workers that leverage threaded libgit2 functions (at the moment, only checkout functionality), and they basically allow the callbacks from the current Worker to be queued to the main JS thread without waiting for the current callback to end.

It is `UNSAFE` because with threaded libgit2 functions there will be a potential risk of deadlock if the callbacks need to lock an object that is already locked by the current Worker.

The `Unsafe` commit in this PR will be reverted once LFS is integrated into nodegit.

### DIFF
--- a/generate/templates/manual/include/thread_pool.h
+++ b/generate/templates/manual/include/thread_pool.h
@@ -8,6 +8,20 @@
 
 #include "async_worker.h"
 
+// Temporary workaround for LFS checkout. Comment added to be reverted.
+// With the threadpool rewrite, a Worker will execute its callbacks with
+// objects temporary unlock (to prevent deadlocks), and we'll wait until
+// the callback is done to lock them back again (to make sure it's thread-safe).
+// LFS checkout lost performance after this, and the proper way to fix it is
+// to integrate nodegit-lfs into nodegit. Until this is implemented, a
+// temporary workaround has been applied, which affects only Workers leveraging
+// threaded libgit2 functions (at the moment only checkout) and does the
+// following:
+// - do not wait for the current callback to end, so that it can send the
+//  next callback to the main JS thread.
+// - do not temporary unlock the objects, since they would be locked back
+//  again before the callback is executed.
+
 namespace nodegit {
   class Context;
   class AsyncContextCleanupHandle;
@@ -17,7 +31,9 @@ namespace nodegit {
     public:
       typedef std::function<void()> Callback;
       typedef std::function<void(Callback, Callback)> QueueCallbackFn;
-      typedef std::function<Callback(QueueCallbackFn, Callback)> OnPostCallbackFn;
+      // Temporary workaround for LFS checkout. Code modified to be reverted.
+      // typedef std::function<Callback(QueueCallbackFn, Callback)> OnPostCallbackFn;
+      typedef std::function<Callback(QueueCallbackFn, Callback, bool)> OnPostCallbackFn;
 
       // Initializes thread pool and spins up the requested number of threads
       // The provided loop will be used for completion callbacks, whenever
--- a/generate/templates/manual/src/async_baton.cc
+++ b/generate/templates/manual/src/async_baton.cc
@@ -46,7 +46,8 @@ namespace nodegit {
       ThreadPool::PostCallbackEvent(
         [jsCallback, cancelCallback](
           ThreadPool::QueueCallbackFn queueCallback,
-          ThreadPool::Callback callbackCompleted
+          ThreadPool::Callback callbackCompleted,
+          bool isThreaded // Temporary workaround for LFS checkout. Code added to be reverted.
         ) -> ThreadPool::Callback {
           queueCallback(jsCallback, cancelCallback);
           callbackCompleted();
@@ -58,13 +59,22 @@ namespace nodegit {
       ThreadPool::PostCallbackEvent(
         [this, jsCallback, cancelCallback](
           ThreadPool::QueueCallbackFn queueCallback,
-          ThreadPool::Callback callbackCompleted
+          ThreadPool::Callback callbackCompleted,
+          bool isThreaded // Temporary workaround for LFS checkout. Code added to be reverted.
         ) -> ThreadPool::Callback {
-          this->onCompletion = callbackCompleted;
+          // Temporary workaround for LFS checkout. Code modified to be reverted.
+          if (!isThreaded) {
+            this->onCompletion = callbackCompleted;
 
-          queueCallback(jsCallback, cancelCallback);
+            queueCallback(jsCallback, cancelCallback);
 
-          return std::bind(&AsyncBaton::SignalCompletion, this);
+            return std::bind(&AsyncBaton::SignalCompletion, this);
+          }
+          else {
+            this->onCompletion = std::bind(&AsyncBaton::SignalCompletion, this);
+            queueCallback(jsCallback, cancelCallback);
+            return []() {};
+          }
         }
       );
 
--- a/generate/templates/manual/src/thread_pool.cc
+++ b/generate/templates/manual/src/thread_pool.cc
@@ -479,10 +479,6 @@ namespace nodegit {
 
       void QueueCallbackOnJSThread(ThreadPool::Callback callback, ThreadPool::Callback cancelCallback, bool isWork);
 
-      static void RunJSThreadCallbacksFromOrchestrator(uv_async_t *handle);
-
-      void RunJSThreadCallbacksFromOrchestrator();
-
       static void RunLoopCallbacks(uv_async_t *handle);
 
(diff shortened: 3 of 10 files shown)


implausible:
@AlexaXs This is not a good move. While it makes sense that we want to fix the regression - this cannot be a permanent solution as it throws lock safety out the window. Basically this allows nodegit to enter a state where the callback calling thread may perform work with a "locked" object, but the lock itself has been released by the callback runner. This is incredibly unsafe and was one of the aims that I intended to fix when updating the library in this way.

This may be a _temporary workaround_ however, we have no idea what the consequences of this change really are other than "it worked before". There definitely existed segfaults around LFS before, and it's very possible that this fix improved stability.

The appropriate change for fixing LFS is not to remove this patch, but to integrate nodegit-lfs into nodegit directly and with a managed solution for working with locks. Frankly, this fixes a performance regression to reintroduce a safety regression... We should aim instead to fix 2 performance regressions without introducing any safety regressions. Nodegit-lfs is poorly designed right now, because it does not use a streaming interface, it loads the entire blob into memory and then sends it to lfs for processing. If we integrated nodegit-lfs directly into nodegit, we could theoretically switch the implementation to:
1) streaming interface
2) no need for callbacks to JS simply for streaming data
3) only using callbacks for authentication

I strongly suggest taking this approach for fixing LFS and looking at a longer term approach that will improve the value of the library. That this is an issue at all should provide enough friction with the team and customers to provide the support you need to build this out at Axosoft.

That all said, if the timeline for fixing lfs properly isn't manageable, we can discuss merging this; however, this PR and associated commits need to be updated to comment out the code and properly document that this was removed as a temporary workaround and not as a "permanent fix". We should label this as an unsafe workaround before it is merged, and we should not remove the code that works as intended.

implausible (changes requested):
@AlexaXs This is not a good move. While it makes sense that we want to fix the regression - this cannot be a permanent solution as it throws lock safety out the window. Basically this allows nodegit to enter a state where the callback calling thread may perform work with a "locked" object, but the lock itself has been released by the callback runner. This is incredibly unsafe and was one of the aims that I intended to fix when updating the library in this way.

This may be a temporary workaround however, we have no idea what the consequences of this change really are other than "it seemed to work before". There definitely existed segfaults around LFS before, and it's very possible that this fix improved stability.

The appropriate change for fixing LFS is not to remove this patch, but to integrate nodegit-lfs into nodegit directly and with a managed solution for working with locks. Frankly, this fixes a performance regression to reintroduce a safety regression... We should aim instead to fix 2 performance regressions without introducing any safety regressions. Nodegit-lfs is poorly designed right now, because it does not use a streaming interface, it loads the entire blob into memory and then sends it to lfs for processing. If we integrated nodegit-lfs directly into nodegit, we could theoretically switch the implementation to:

streaming interface
no need for callbacks to JS simply for streaming data
only using callbacks for authentication
I strongly suggest taking this approach for fixing LFS and looking at a longer term approach that will improve the value of the library. That this is an issue at all should provide enough friction with the team and customers to provide the support you need to build this out at Axosoft.

That all said, if the timeline for fixing lfs properly isn't manageable, we can discuss merging this; however, this PR and associated commits need to be updated to comment out the code and properly document that this was removed as a temporary workaround and not as a "permanent fix". We should label this as an unsafe workaround before it is merged, and we should not remove the code that works as intended.

AlexaXs:
@implausible, first of all than you so much for reviewing this PR, and for the tremendous work you did to rewrite the threadpool. Also thanks a lot for your comments on the approach to fix LFS properly.

Obviously I hadn't understood all the flow involved in the threadpool with thread-safety, so apologies for the unsafe changes in the PR.

I've checked previous PRs related with the threadpool and thread-safety to understand better the present code:

- https://github.com/nodegit/nodegit/issues/827
- https://github.com/nodegit/nodegit/pull/836
- https://github.com/nodegit/nodegit/pull/1019

I'd appreciate if you could confirm what I understand about the present code:

1. [This temporaryUnlock](https://github.com/nodegit/nodegit/blob/bae468cf3da6ae5ef6d3f89d11bf89f83e8db94a/generate/templates/manual/src/thread_pool.cc#L381) is here to ***prevent a deadlock*** in the main JS thread in this situation: a Worker execution (asynchronous) has locked some libgit2 objects (only of type `git_repository` and `git_index` at the moment), and while locked, the main JS thread runs a synchrounous NodeGit API that could potentially also lock the same objects.

2. [This waiting](https://github.com/nodegit/nodegit/blob/bae468cf3da6ae5ef6d3f89d11bf89f83e8db94a/generate/templates/manual/src/thread_pool.cc#L394) is what makes NodeGit remain ***thread-safe*** while the JS callback is running and the locked objects are temporary unlocked.

If my understanding on those points are correct, it'd be great if you could review this change to the PR consisting on ***commenting out the temporaryUnlock*** in the first point above. Let me know if my thoughts are correct concerning this possible update to the PR:

3. `LockMaster` locks all the objects at once, and that already prevents deadlocks between async executions.
4. At the moment, [synchronous NodeGit APIs never lock any object](https://github.com/nodegit/nodegit/blob/bae468cf3da6ae5ef6d3f89d11bf89f83e8db94a/generate/templates/partials/sync_function.cc#L38).

Unless there's something I'm not seeing, by commenting out the temporaryUnlock, the objects initially locked by the async Worker ***will remain locked*** until [the COMPLETED event is received](https://github.com/nodegit/nodegit/blob/bae468cf3da6ae5ef6d3f89d11bf89f83e8db94a/generate/templates/manual/src/thread_pool.cc#L371) (when [the Worker finishes with the async execution](https://github.com/nodegit/nodegit/blob/bae468cf3da6ae5ef6d3f89d11bf89f83e8db94a/generate/templates/manual/src/thread_pool.cc#L179)), keeping NodeGit thread-safe.

As I see it, with this change, although we open the possibility to suffer from deadlocks in the future, ***with the present code*** NodeGit remains thread-safe and there is no possibility of deadlocks.

If you think these thoughts are correct and ***commenting out the temporaryUnlock*** is a valid temporal workaround, it'd be great if you could merge this PR ***until LFS is fixed properly***. In that case I'd update this PR this way:

- I'll re-label the PR as `Unsafe temporal workaround: LFS checkout performance regression`.
- I'll add a new commit to put back the removed code and will comment it out instead, documenting the reason for the temporal workaround.

Let me know please if everything makes sense. And happy 2022 by the way! 😄

AlexaXs:
@implausible, I've modified the description and the commits of this PR to see if they could work as an `UNSAFE` temporary workaround until LFS is integrated into NodeGit:

- The changes will be limited to callbacks from Workers leveraging threaded libgit2 functions (at the moment only `checkout` functionality). The rest of the functionality won't be affected.
- The changes will allow callbacks to be queued to the main JS thread without waiting for the current one to end, and will not apply a `temporaryUnlock`, since it would lock back again before the callback is executed. This way we make sure that the callback calling thread will perform work with objects locked.
- The changes will be `UNSAFE` because the callbacks might need to lock objects already locked by the current Worker, ending on a `deadlock`. Tests have been added to check this situation in two different scenarios, with `threaded` and `non-threaded` libgit2 functions. The test `can run async callback on checkout without deadlocking` have been added and skipped, since it will fail with these changes, pointing to why these changes are `UNSAFE`. The behaviour here is the same as with the threadpool before the rewrite.
- There are comments starting with `Temporary workaround for LFS checkout` in the code related to the changes.

We'll be working on the integration of LFS into nodegit, that will revert this `Unsafe` commit if you think it can be merged as a temporary workaround.

implausible:
I think that generally covers my major concerns. I would be open to this workaround since it is targeted and addresses real customer impact. I am excited to see LFS become nodegit native in the future!

implausible:
Would you kindly add a test in https://github.com/nodegit/nodegit/blob/master/test/tests/worker.js to demonstrate that this code is safe during shutdowns?

AlexaXs:
> I would be open to this workaround since it is targeted and addresses real customer impact. I am excited to see LFS become nodegit native in the future!

Thank you so much @implausible, and yes, I am excited too about the LFS integration!

Sorry for the delay in pushing the tests. I`ve added a couple of tests, the first one terminates a worker thread that is running checkouts in a loop where a git filter is used, to demonstrate that it is terminated gracefully. The second one checks that the objects involved in multiple checkouts are tracked and will be freed when the worker thread is terminated, preventing memory leaks.

implausible (approved):

Issue #626, opened 2015-06-27
Electron - Module did not self-register 


ybouhjira:
I get this error when I try to `require` nodegit (version 0.4.1) in electron (version 0.28.3)

```
A JavaScript error occured in the browser process
Uncaught Exception:
Error: Module did not self-register.
    at Error (native)
    at Object.module.(anonymous function) (ATOM_SHELL_ASAR.js:137:20)
    at Object.module.(anonymous function) [as .node] (ATOM_SHELL_ASAR.js:137:20)
    at Module.load (module.js:351:32)
    at Function.Module._load (module.js:306:12)
    at Module.require (module.js:361:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/home/youssef/Desktop/test/node_modules/nodegit/lib/nodegit.js:8:12)
    at Module._compile (module.js:426:26)
    at Object.Module._extensions..js (module.js:444:10)
```
I added `engines.electron` in package.json BTW 
```
  "engines": {
    "electron": "0.28.3"
  },
```

johnhaley81:
Try adding the module as `atom-shell`

ybouhjira:
I tried and I got the same error.

maxkorp:
What's your build process? Can you walk me through how it's all built? Electron build processes are a bit odd compared to general node stuff, but I'm pretty familiar.

ybouhjira:
I just install nodegit using `npm install` and it says in the output that it's compiling it. Is that enough

jhansen-tt:
I'm seeing the same error.  My 'npm install' appears to be building the nodegit.node binary from scratch because it can't find one for x64 linux on S3.  Not sure if the nodegit.node file should be runnable, but when I try to run it I get a seg fault crash, and gdb isn't helpful on giving any more information.  Any chance you can post an x64 Linux binary to S3, for Ubuntu Linux?  Anything I can help with on debugging this?  For now I have to revert back to 0.3.3.

My nodegit works great on Mac but that's because there's a binary out there on S3, but my custom-built binary for Linux is crashing.

maxkorp:
Just to verify, you're doing the build AFTER setting the engines field? That only has an effect at compile time, not at build time. 

The .node file should be requirable. If you try requiring it in a normal node context, does that work?

jhansen-tt:
[...]

Issue #536, opened 2015-04-10
Unable to compile or install with npm install nodegit


weikinhuang:
I am trying to get nodegit installed on windows (7, 8.1 x64) and iojs 1.6.x. I am getting the following errors. And I've tried the different build options specified in the docs. I am able to compile other modules such as `sqlite` without issues.

```
> nodegit@0.3.3 install D:\Documents\Projects\node_modules\nodegit
> node lifecycleScripts/install

[nodegit] Fetching binary from S3.
[nodegit] Failed to install prebuilt binary, building manually.
[nodegit] Making sure dependencies are available and native code is generated
[nodegit] Everything is ready to go, attempting compilation
[nodegit] Building native node module.
{ [Error: Command failed: C:\Windows\system32\cmd.exe /s /c "D:\Documents\Projects\node_modules\nodegit\node_modules\.bin\pangyp rebuild"
gyp ERR! build error 
gyp ERR! stack Error: `C:\Program Files (x86)\MSBuild\12.0\bin\msbuild.exe` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onExit (D:\Documents\Projects\node_modules\nodegit\node_modules\pangyp\lib\build.js:271:23)
gyp ERR! stack     at emitTwo (events.js:87:13)
gyp ERR! stack     at ChildProcess.emit (events.js:169:7)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:1044:12)
gyp ERR! System Windows_NT 6.3.9600
gyp ERR! command "C:\\Program Files\\iojs\\node.exe" "D:\\Documents\\Projects\\node_modules\\nodegit\\node_modules\\pangyp\\bin\\node-gyp.js" "rebuild"
gyp ERR! cwd D:\Documents\Projects\node_modules\nodegit
gyp ERR! node -v v1.6.4
gyp ERR! pangyp -v v2.1.0
gyp ERR! not ok 
]
  killed: false,
  code: 1,
  signal: null,
  cmd: 'C:\\Windows\\system32\\cmd.exe /s /c "D:\\Documents\\Projects\\node_modules\\nodegit\\node_modules\\.bin\\pangyp rebuild"' }
undefined

> nodegit@0.3.3 postinstall D:\Documents\Projects\node_modules\nodegit
> node lifecycleScripts/clean

[nodegit] Cleaning up after ourselves.
[nodegit] Deleting unneeded files.
[nodegit] Finished deleting files.
[...]

PR #1888, opened 2022-01-11
RSA SHA2 256/512 key upgrade support RFC 8332 #536 (#626)

julianmesa-gitkraken:
Notes:
* Host Key RSA 256/512 support [#536](https://github.com/libssh2/libssh2/issues/536)
* Client side key hash upgrading for RFC 8332
* Support for server-sig-algs, ext-info-c server messages
* Customizing preferred server-sig-algs via the preference LIBSSH2_METHOD_SIGN_ALGO

Credit: Anders Borum, Will Cosgrove

This PR will fix the error when trying to use ssh protocol in github due this changes:
[Improving Git protocol security on GitHub](https://github.blog/2021-09-01-improving-git-protocol-security-github/)

Original commit: [RSA SHA2 256/512 key upgrade support RFC 8332 #536 (#626)](https://github.com/libssh2/libssh2/commit/64a555d6f5aafed504a10e5b756e85c91b1d56ce)

Also more commits from libssh2 has been added:

- Fix a memcmp errors in code that was changed from memmem to memcmp
- ssh: Add support for userauth banner
- Support rsa-sha2 agent flags
- Legacy Agent support for rsa2 key upgrading/downgrading
- free RSA2 related memory
- NULL terminate server_sign_algorithms string


test/tests/clone.js
```diff
@@ -286,6 +289,66 @@ describe("Clone", function() {
...
+      };
+
+      try {
+        await exec("taskkill /im pageant.exe /f /t");
+      } catch (e) {
+        try {
+          await exec("taskkill /im pageant_sha1.exe /f /t");
+        } catch(e) {}
+        try {
+          await exec(pageant, [privateKey]);
```
ianhattendorf, line 318: It looks like elsewhere we ```await exec(`powershell -command "Start-Process ${old_pageant} ${privateKey}`);```, should we be doing that here as well? Or should we be doing this in the other locations?

julianmesa-gitkraken: my bad, I will fix it

ianhattendorf (approved):

PR #1894, opened 2022-02-23
Upgrade libgit2 to 1.3.1

zawata:
Several changes were made:

 - commit signing callback was deprecated in favor of a commit creation callback
    - callback returns an OID through a pointer to an oid struct. not sure if this could be handled better, looking for suggestions
    - contains a list parameter that required some modification to accommodate
 - public git_buf modification functions were deprecated as modifying git_buffers are now considered internal only
    - the only place this affects us is in the git_filter apply callback function which has also been deprecated but will be modified in another PR since it requires bigger changes
 - adjusted arrayType regex to accommodate array parameters
 - removed win32 long-paths since it's included in libgit2 config handling
 - fix compilation of new ntlmclient version
 
 Note: A majority of the PR changes are the new Libgit2-docs.json


generate/input/descriptor.json
```diff
@@ -1667,6 +1667,20 @@
           "return": {
             "isResultOrError": true
           }
+        },
+        "git_graph_reachable_from_any": {
```
ianhattendorf, line 1671: Can we get a test for this?

ianhattendorf (changes requested):
Looking good, just a few things to address.

test/tests/clone.js
```diff
@@ -84,54 +80,6 @@ describe("Clone", function() {
...
-          callbacks: {
-            certificateCheck: () => 0
-        }
-      }
-    };
-
-    fse.ensureDirSync(longClonePath);
-
-    if (process.platform === "win32") {
-      NodeGit.Libgit2.opts(NodeGit.Libgit2.OPT.SET_WINDOWS_LONGPATHS, 1);
```
ianhattendorf, line 101: Now that we're moving to the updated longpaths support, we should make sure to update the tests to account for the changes instead of just removing them. See https://github.com/libgit2/libgit2/pull/5823 for the expected behavior.

AlexaXs (commented):
Changes in `vendor/libgit2` look good. Merged changes onto libgit2-1.3.0 maintain the original code semantics.

generate/templates/partials/configurable_callbacks.cc
```diff
@@ -1,3 +1,4 @@
+#include <git2/oid.h>
```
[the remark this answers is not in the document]
zawata: valid point. I didn't really like the solution of needing to make it specific to OIDs in the first place but I couldn't find anything better at the time. I'll give it another look to see if it can be made generic but if not I'll take this suggestion.

generate/input/descriptor.json
zawata: turns out this function was totally broken. wouldn't have known without trying to test it :sweat_smile:

generate/templates/partials/configurable_callbacks.cc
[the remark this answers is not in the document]
zawata: I couldn't come up with anything better. maybe something to look into later.
the header file was already being included by the dependency inclusion template in struct_content.

test/tests/rebase.js
```diff
@@ -1891,13 +1891,13 @@ describe("Rebase", function() {
         .then(function(commit) {
           // verify that the "ours" branch has moved to the correct place
           assert.equal(commit.id().toString(),
-          "24250fe6bd8a782ec1aaca8b2c9a2456a90517ed");
+          "617cd03370dd799f372e9dcfcd0b097aede1bd7f");
```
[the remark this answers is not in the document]
zawata: I'm not actually sure... I didn't actually touch anything related to it or modify any of the commits and the new shas are consistently. even passing on CI. I could take another look if it's significant.

test/tests/clone.js
zawata: New tests have been implemented. they were converted from cloning tests to checkout tests to accommodate the slight change in functionality

wmertens:
In the meantime libgit 1.4.2 is out, would that be an easy upgrade?

test/tests/rebase.js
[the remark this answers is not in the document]
ianhattendorf: Yes, I would take a look. This means that something changed with how libgit2 creates a commit, since the sha is generated from the content + header.

zawata:
@wmertens when I started this 1.4 had just come out and we decided to make changes incrememental rather than large(r) version jumps to help with API stabilization and minimize implementation problems. I do have the 1.4 upgrade in my backlog but I think theres some higher priority tasks to get done first(at least from my end) like filter streaming.

The actual upgrade to 1.4 doesn't look super complex but i wouldn't no for certain until I start working on it.

test/tests/repository.js
```diff
@@ -362,15 +362,15 @@ describe("Repository", function() {
...
     .then(function(analysisReport) {
 
-      assert.equal(analysisReport.repositorySize.commits.count, 992);
-      assert.equal(analysisReport.repositorySize.commits.size, 265544);
-      assert.equal(analysisReport.repositorySize.trees.count, 2416);
-      assert.equal(analysisReport.repositorySize.trees.size, 1188325);
-      assert.equal(analysisReport.repositorySize.trees.entries, 32571);
-      assert.equal(analysisReport.repositorySize.blobs.count, 4149);
-      assert.equal(analysisReport.repositorySize.blobs.size, 48489622);
+      assert.equal(analysisReport.repositorySize.commits.count, 993);
```
ianhattendorf, line 365: Oops, just realized we use the `nodegit/test` repo here when we should have used a separate repo that's frozen.

@implausible would you mind creating a `nodegit/test-frozen` repo when you have a minute?

Then I'll push up the old version of `nodegit/test` to that repo so we have a consistent repo for repo analysis and don't needlessly break tests for older commits each time we update the test repo.

zawata:
In light of the recent libgit2 security release, I'll update this PR to use libgit 1.3.1. Luckily this won't require any code or libgit2-docs changes. I've already confirmed it builds.

@ianhattendorf I've updated my fork of the libgit-next branch. Please push it when convenient and I'll update the submodule in this PR. https://github.com/zawata/libgit2/tree/libgit-next

zawata:
PR updated to 1.3.1

test/tests/repository.js
ianhattendorf: @zawata I created https://github.com/nodegit/test-frozen, go ahead and swap the const repo over to that. I also deleted the longpath branch from this PR so the numbers will change. Hopefully they're the same as they were before, but if not it's not the end of the world.

test/tests/rebase.js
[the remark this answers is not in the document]
ianhattendorf: Looks like these commits changed because `createCommitWithSignature` and `rebase.commit` now have the same newline handling. Verified that commits created before this PR, with this PR, and via the CLI are create valid signatures, and that NodeGit before and with this PR and the CLI are able to successfully verify the signatures.

ianhattendorf (approved):

PR #1897, opened 2022-03-11
Update windows 2016 CI to latest

julianmesa-gitkraken:
To update Windows box we need to upgrade node-gyp, so we need python 3.6 in ubuntu 16.04, but last version is 3.5. This is the reason Python 3.6 is build from source.


ianhattendorf (changes requested):
- Make sure to commit your `package-lock.json` changes
- There's a breaking change in @mapbox/node-pre-gyp v1.0.0 we need to handle, see https://github.com/mapbox/node-pre-gyp/blob/master/CHANGELOG.md (related: #1866)

ianhattendorf (approved):

PR #1900, opened 2022-03-21
Fix electron-npg-automator

julianmesa-gitkraken:
After node-gyp upgrade electron-npg-automator is not working.
Downgrade node-gyp to 8.4.1
Removed babel stuff, there is a conflict with fsevents package in electron-npg-automator


ianhattendorf:
Fixed in #1901

PR #1903, opened 2022-03-23
Expose git_push_negotiation callback

zawata:
For use in verifying or rejecting the push ref-update list before pushing.


zawata:
On hold until #1894 is merged.

PR #1905, opened 2022-03-31
Statically compile OpenSSL on linux for electron

zawata:
Several fixes included, they should be documented in the code.

Notable Changes:
 - a code patching system
   - only used to fix openssl old linux OS's(GLIBC <= 2.25) for now but can be expanded in the future if necessary
 - openssl configure and build support for linux
   - notably there is a flag passed here to hide all openssl symbols, without this, nodegit would segfault in cases where the electron environment has loaded in openssl
  - Apparently we were building and shipping some node binaries without code in them, this was an artifact of making action-only targets but not declaring them as action-only. I fixed that.
  - Also configured openssl on linux and mac to only build and install the headers and libs. we no longer build the additional apps, fuzzers, and docs. should result in a small speed up

### DIFF
--- a/generate/templates/templates/binding.gyp
+++ b/generate/templates/templates/binding.gyp
@@ -9,6 +9,7 @@
   "targets": [
     {
       "target_name": "acquireOpenSSL",
+        "type": "none",
         "conditions": [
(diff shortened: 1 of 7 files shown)


julianmesa-gitkraken (approved):
Great work

ianhattendorf (approved):

PR #1928, opened 2022-08-10
Preemptively Patch OpenSSL 1.1.1q

gitkraken-jacobw:
This adds a patch for MacOS that allows OpenSSL tests to compile for that operating system.

Test compilation was broken in https://github.com/openssl/openssl/commit/a1d80edcf830739131e0567dc03b1e80b7988b1e and fixed in https://github.com/openssl/openssl/commit/60f011f584d80447e86cae1d1bd3ae24bc13235b. The patch introduces the fix.

I was not able to build for Windows, so I have not yet tested if the updated patching system works there.

This is meant for when NodeGit is bumped to 1.1.1q.


ianhattendorf:
> I was not able to build for Windows, so I have not yet tested if the updated patching system works there.

My guess is this won't work, at least on a default Windows installation. It might work if users installed git by adding the unix tools to their path, however not all users will do that. Most likely if we want to apply patches on Windows we'd need to find a JS based approach (e.g. npm package). For now, we can probably just not apply the patch on Windows as IIRC OpenSSL 1.1.1q builds fine on that.

gitkraken-jacobw:
> > I was not able to build for Windows, so I have not yet tested if the updated patching system works there.
> 
> My guess is this won't work, at least on a default Windows installation. It might work if users installed git by adding the unix tools to their path, however not all users will do that. Most likely if we want to apply patches on Windows we'd need to find a JS based approach (e.g. npm package). For now, we can probably just not apply the patch on Windows as IIRC OpenSSL 1.1.1q builds fine on that.

Alright, sounds good. Pushed up a change that removes Windows patching.

utils/acquireOpenSSL.js
```diff
@@ -42,10 +42,10 @@ class HashVerify extends stream.Transform {
 }
 
 // currently this only needs to be done on linux
-const applyOpenSSLPatches = async (buildCwd) => {
+const applyOpenSSLPatches = async (buildCwd, operatingSystem) => {
   try {
     for (const patchFilename of await fse.readdir(opensslPatchPath)) {
-      if (patchFilename.split(".").pop() === "patch") {
+      if (patchFilename.split(".").pop() === "patch" && patchFilename.split("-")[1] === operatingSystem) {
```
ianhattendorf, line 48: What do you think about changing this so we can specify `all` for the OS on a patch? Looking at the new patch this is probably something we should include on all OSes anyway.

gitkraken-jacobw: I think this is a good idea. I'll switch the patch over to the new designation as well.

PR #1932, opened 2022-09-08
Work around apparent SSL incompatibility with Node 18

chuckdries:
When attempting to install under node 18 I encounter this error message, 
```
/Users/chuckdries/.node-gyp/18.0.0/include/node/openssl/macros.h:155:4: error: "OPENSSL_API_COMPAT expresses an impossible API compatibility level"
#  error "OPENSSL_API_COMPAT expresses an impossible API compatibility level"
   ^
1 error generated.
```

so I applied the solution from this stackoverflow and it worked. ¯\\\_( ツ )_/¯  https://stackoverflow.com/questions/70541687/linking-local-openssl-to-nodejs-c-addon


ianhattendorf:
Hmm, probably related to the move to OpenSSL 3. I'm worried this will break for older node versions, let's see what happens in CI.

ianhattendorf:
@chuckdries looks like CI passed, do you want to try adding `18` to `jobs.(linux,macos)-tests.strategy.matrix.node` inside `.github/workflows/tests.yml` so we can see if we pass there?

zawata:
I don't think this is relevant since no openssl 1.1 nodejs version is in active support and we switched to openssl 3

Issue #1935, opened 2022-09-20
repo.getReferenceNames( nodegit.Reference.TYPE.DIRECT ) does not work.


axkibe:
## System information
  - node version: v16.14.2
  - npm or yarn version: 8.5.0
  - OS/version/architecture: debian
  - Applicable nodegit version: 0.27.0 (or HEAD)

## Issue
const refNames = await repo.getReferenceNames( nodegit.Reference.TYPE.DIRECT );

will return an empty array. TYPE.ALL will work.

## Solution:
https://github.com/nodegit/nodegit/blob/master/lib/repository.js#L343

is missing a call.

it should be:
```
      return type === Reference.TYPE.ALL || reference.type( ) === type;
```
instead of
```
      return type === Reference.TYPE.ALL || reference.type === type;
```

ianhattendorf:
Fixed by #1942.

PR #1942, opened 2022-10-13
type is a call

axkibe:
as per issue #1935


zawata:
huh. this has been busted for 4 years?

seems to be introduced in this commit:
https://github.com/nodegit/nodegit/commit/adcc5c520aae1a3351d9c463ee0ae7a207e666d1

axkibe:
It's a big library, and I guess not all functionality is constantly used, since the ALL worked I guess this is what most used and no one wanted DIRECT only. I only stumbled over this while doing some experiments, not sure if I'm not going for ALL anyway at the end.

ianhattendorf (approved):
Thanks!

PR #1943, opened 2022-10-17
Fix Alloc-Dealloc mismatches

zawata:
An Alloc-Dealloc Mismatch causes undefined behavior. Kind of a minor problem but should be fixed nonetheless.

for more details
https://stackoverflow.com/a/10854437/4776273

### DIFF
--- a/generate/templates/partials/async_function.cc
+++ b/generate/templates/partials/async_function.cc
@@ -56,7 +56,7 @@ NAN_METHOD({{ cppClassName }}::{{ cppFunctionName }}) {
         {% if arg.cppClassName == 'Array' %}
           {
             v8::Local<v8::Array> tempArray = v8::Local<v8::Array>::Cast(info[{{ arg.jsArg }}]);
-            baton->{{ arg.name }} = new {{ arg.cType|unPointer }}[tempArray->Length()];
+            baton->{{ arg.name }} = ({{ arg.cType|unPointer }}*)malloc(sizeof({{ arg.cType|unPointer }}) * tempArray->Length());
             for (uint32_t i = 0; i < tempArray->Length(); ++i) {
               auto conversionResult = Configurable{{ arg.arrayElementCppClassName }}::fromJavascript(
                 nodegitContext,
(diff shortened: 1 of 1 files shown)


generate/templates/manual/revwalk/file_history_walk.cc
```diff
@@ -31,6 +31,14 @@ class FileHistoryEvent {
     if (commit != NULL) {
       git_commit_free(commit);
     }
+
+    if(from != NULL) {
```
ianhattendorf, line 35: Don't need to null check before free.

generate/templates/partials/async_function.cc
```diff
@@ -56,20 +56,21 @@ NAN_METHOD({{ cppClassName }}::{{ cppFunctionName }}) {
...
+            baton->{{ arg.name }} = ({{ arg.cType|unPointer }}*)malloc(sizeof({{ arg.cType|unPointer }}) * tempArray->Length());
             for (uint32_t i = 0; i < tempArray->Length(); ++i) {
               auto conversionResult = Configurable{{ arg.arrayElementCppClassName }}::fromJavascript(
                 nodegitContext,
                 Nan::Get(tempArray, i).ToLocalChecked()
               );
 
               if (!conversionResult.result) {
-                delete[] baton->{{ arg.name }};
+                // TODO free previously allocated memory
```
ianhattendorf, line 67: Does this still need to be handled?

generate/templates/manual/revwalk/file_history_walk.cc
ianhattendorf: Although I guess we're already doing this everywhere :shrug:

generate/templates/partials/async_function.cc
zawata: @julianmesa-gitkraken this commit came from you.

julianmesa-gitkraken: Yes but I will do in another PR. I need to store all allocated memory in a map, then if I get an error reading javascript arguments we have to release all allocated memory, then throw the error.

PR #1945, opened 2022-10-19
Fix tag createWithSignature function

julianmesa-gitkraken:
Fix tag createWithSignature function
Also tests fixed

### DIFF
--- a/lib/tag.js
+++ b/lib/tag.js
@@ -66,7 +66,7 @@ const deprecatedCreateWithSignatureHelper = util.deprecate(function(repo, oidTar
  *                                   representing the signed message
  * @return {Oid}
  */
-Tag.createWithSignature = function(
+Tag.createWithSignature = async (
   repo,
   tagName,
   target,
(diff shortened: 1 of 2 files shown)


lib/tag.js
```diff
@@ -71,7 +71,7 @@ Tag.createWithSignature = function(
   signingCallback
 ) {
   let tagBuffer;
-  return Tag.createBuffer(repo, tagName, target, tagger, message)
+  return Tag.createBuffer(repo, tagName, target.id(), tagger, message)
```
ianhattendorf, line 74: Can we check if `target.id` exists, and if not throw a deprecation warning while still functioning? That way we give users a heads up that they're using the API incorrectly without abruptly breaking things for them. Something like (pseudocode, not tested)
```
const deprecatedCreateWithSignatureHelper = util.deprecate(function(oidTarget) {
  return repo.getCommit(oidTarget);
}, "Tag.createWithSignature target should be a Git Object, not Oid");

Tag.createWithSignature = async function(
  ...


  if (!target.id) {
   target = await deprecatedCreateWithSignatureHelper(target);
  }

...  
```

See https://nodejs.org/api/util.html#utildeprecatefn-msg-code if you aren't familiar with `util.deprecate`, we use this elsewhere in NodeGit.

ianhattendorf (changes requested):

lib/tag.js
julianmesa-gitkraken: Done.

julianmesa-gitkraken:
@ianhattendorf Changed `createWithSignature` using async/await.
Also added a test to show the deprecation warning.

ianhattendorf (approved):
Thanks for adding the deprecation message :+1:

PR #1948, opened 2022-10-28
Partial stash support

ianhattendorf:
Bring in https://github.com/libgit2/libgit2/pull/6330


zawata:
compared to the original https://github.com/Axosoft/nodegit/pull/1, looks fine to me.

PR #1951, opened 2022-11-02
Fix intermittent crash on Electron 18+

zawata:
This was *not* fun to track down. Probably ~3 weeks of work?

This PR bumps nan to a custom version hosted on the axosoft org(could be moved to nodegit if necessary) and bumps the internal field count on object templates to account for the updates made to nan.

more details on what is happening and why, here:
https://github.com/Axosoft/nan/pull/1
and here:
https://github.com/electron/electron/blob/693cf0ebdebb272494610a5ed2e9c405e825813e/patches/node/be_compatible_with_cppgc.patch


zawata:
CI failures appear to be caused by an npm bug so ill publish the axosoft nan package and put it in this PR.

ianhattendorf (approved):
Nice work :tada:

PR #1961, opened 2022-12-15
Bump got to latest version to fix vulnerability

vinodkumarsharma276:
We are consuming nodegit in our project transitively its dependency -- got package  before 11.8.5 and 12.1.0 for Node.js allows a redirect to a UNIX socket. 
Please approve this PR and push a new version to npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "@axosoft/nan": "^2.18.0-gk.1",
     "@mapbox/node-pre-gyp": "^1.0.8",
     "fs-extra": "^7.0.0",
-    "got": "^10.7.0",
+    "got": "^12.5.3",
     "json5": "^2.1.0",
     "lodash": "^4.17.14",
     "node-gyp": "^9.3.0",


vinodkumarsharma276:
@johnhaley81 @tbranyen @implausible

tbranyen:
@vinodkumarsharma276 there has not been a NodeGit release in over 2 years. So it is unlikely this will be landed and a new release cut anytime soon. I would suggest using a packages override or finding another way of getting this patch into your application.

@implausible is Axosoft still actively supporting NodeGit? I don't want to start coming in and merging and closing PRs/issues and cause problems. Let me know here or privately so we can figure out how to get easy wins like this landed timely.

johnhaley81:
I think @implausible is no longer at Axosoft (which is now called GitKraken). @Mr-Wallet is still there though and he might be able to point this in the right direction.

zawata:
I don't have permissions to merge PRs but I am on the team that supports this project. @ianhattendorf is currently the only active axosoft employee who can merge PRs in this repository

Unfortunately every other member of my team is out this week(including Ian) is out on vacation so I can't merge this. I believe Ian will be back Monday and we can work on getting this merged and updated. In the mean-time I agree that package overrides will be your best bet. Something like:

```
  "resolutions": {
    "**/got":"^12.5.3"
  }
```

As for current/future support, obviously things move a little slow in this repo with the other work we're assigned internally but I don't believe we have any plans to abandon this project. I know that we were also waiting until after the new year to merge a couple large PRs that we wanted lots of testing on and we've been wanting to do some overhauls on our build pipeline and our release process but we haven't had the time to implement any ideas.

weedz:
I've experimented a bit with updating node modules for nodegit and discovered that `got@12.x` is ESM only. From their README (https://www.npmjs.com/package/got):

> Warning: This package is native [ESM](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) and no longer provides a CommonJS export. If your project uses CommonJS, you'll have to [convert to ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) or use the [dynamic import()](https://v8.dev/features/dynamic-import) function. Please don't open issues for questions regarding CommonJS / ESM. You can also use [Got v11](https://github.com/sindresorhus/got/tree/v11.8.3) instead which is pretty stable. We will backport security fixes to v11 for the foreseeable future.

So probably won't work with just this change (but should be fine with `got@11`).

ianhattendorf:
I've bumped got to the latest 11.x in #1969 (along with updated a few other locked packages).

A stable release is still planned for the near future, other tasks just keep taking priority unfortunately and I've had zero time outside of work recently.

PR #1993, opened 2024-03-03
Added Arm64 support for Windows and Linux

gpetrov:
* improved OpenSSL download and compile on Windows
* increased the default Visual Studio Build Tools to 2019 as this is the default installed also with NodeJS 18+ and also includes arm64 support
* increased the OpenSSL version to 1.1.1w for arm64 support
* implemented arm64 compile support for Windows in acquireOpenSSL.js and libgit2.gyp
* implemented arm64 compile support for Linux with arm64

Can you check @ianhattendorf

### DIFF
--- a/utils/acquireOpenSSL.js
+++ b/utils/acquireOpenSSL.js
@@ -16,7 +16,7 @@ const pipeline = promisify(stream.pipeline);
 
 const packageJson = require('../package.json')
 
-const OPENSSL_VERSION = "1.1.1t";
+const OPENSSL_VERSION = "1.1.1w";
 const win32BatPath = path.join(__dirname, "build-openssl.bat");
 const vendorPath = path.resolve(__dirname, "..", "vendor");
 const opensslPatchPath = path.join(vendorPath, "patches", "openssl");
@@ -118,7 +118,7 @@ const buildDarwin = async (buildCwd, macOsDeploymentTarget) => {
 
 const buildLinux = async (buildCwd) => {
   const arguments = [
-    "linux-x86_64",
+    process.arch === "x64" ? "linux-x86_64" : "linux-aarch64",
     // Electron(at least on centos7) imports the libcups library at runtime, which has a
     // dependency on the system libssl/libcrypto which causes symbol conflicts and segfaults.
     // To fix this we need to hide all the openssl symbols to prevent them from being overridden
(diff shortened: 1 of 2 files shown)


ianhattendorf:
Thank you! I can't make any promises when I'll be able to look at this but I do want to add support for this.

gpetrov:
@zawata maybe you can add this one as well, when you are bumping nan also

zawata:
@gpetrov It's out of scope right now unfortunately. I have a couple comments on your approach here and outside of this PR, https://github.com/nodegit/electron-npg-automator will need to be updated as well. I don't think arm64 runners are available outside of enterprise orgs right now, but I'd love to research that as well(maybe GK can help somehow?).

That being said we've been researching support for arm64 builds internally, Nodegit is absolutely a blocker for that so we'll be able utilize this work.

My comments:
 - I'd like to make clearer the separation of "targetArch" and "hostArch" rather than just checking `process.arch` and making assumptions
 - Visual Studio 2019 implements support for arm64 cross-compilation but 2022 is required for arm64-native compilation. I updated the Visual-Stdio installation selection to determine what node-gyp detects but I haven't test it for cross-compilation
 
I sketched out some of these changes when I was researching windows-arm64 a couple weeks ago, but they're not tested for anything other-than windows-arm64 hosts.
https://github.com/zawata/nodegit/commit/4b4a17e9d6143450d701a22d92b869af41b1a343

gpetrov:
Thanks @zawata for the feedback.

You can indeed finetune the VS 2022 support.

I was able to fully build the windows Arm64 nodegit binaries and deploy them.

The build was done on an windows 11 Arm64 VM hosted on my MacBook Pro with M1pro ran by UTM https://mac.getutm.app

Also there are windows Arm64 runners available:

https://github.com/actions/virtual-environments/issues/768

zawata:
> windows 11 Arm64 VM hosted on my MacBook Pro with M1pro ran by UTM

This is also how I tested it. I was under the assumption that Visual Studio 2022 was required for native-arm64 builds but if I'm mistaken then vs2019 is a better minimum. I'm still hoping we can have better version selection than hard-coding the path to 2019 and falling back to a non-standard npm-config item.

> Also there are windows Arm64 runners available:

Yes, but they're only available as dedicated larger-runners. Open-source projects like nodegit can't configure windows and linux arm64 runners yet.

gpetrov:
It's better to have visual studio 2022 for best Arm64 support, see:

https://learn.microsoft.com/en-us/visualstudio/install/visual-studio-on-arm-devices?view=vs-2022

gpetrov:
@zawata could you please check also this PR - it is a good addition as well now that you are pushing some updates.

zawata:
superseded by #2028

PR #1995, opened 2024-04-09
Don't free the given repo on error in getReferences and getRemotes

zawata:
This fixes a crash that occurs on windows when trying to run `Repository.getReferences()` on a repo with refs that exceed the path-length limit in libgit2 more than once.

I don't know how else to trigger this issue so I can't really add a test for it.

### DIFF
--- a/generate/templates/manual/repository/get_remotes.cc
+++ b/generate/templates/manual/repository/get_remotes.cc
@@ -51,6 +51,8 @@ void GitRepository::GetRemotesWorker::Execute()
     if (giterr_last() != NULL) {
       baton->error = git_error_dup(giterr_last());
     }
+
+    git_repository_free(repo);
     delete baton->out;
     baton->out = NULL;
     return;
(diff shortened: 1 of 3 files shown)


generate/templates/manual/repository/get_remotes.cc
```diff
@@ -74,7 +74,6 @@ void GitRepository::GetRemotesWorker::Execute()
       }
 
       git_strarray_free(&remote_names);
-      git_repository_free(repo);
```
ianhattendorf, line 77: This free should be fine unless I'm missing something. `repo` was opened inside this function, `baton->repo` was opened outside this function.

It actually should probably be freed in the above returns as well...

ianhattendorf: It looks like opening a new repo was to reduce the amount of time we need to lock (just lock while getting the repo path from the repo object vs. lock for the entire function).

You can see in `get_references.cc` in `AcquireLocks()` we pass `baton->repo` to the `LockMaster` so it will be locked while `Execute()` is running vs. in `get_remotes.cc` it isn't, and instead a new `LockMaster` is constructed just while we're opening a new repo instance (this could actually be shortened to just getting the path).

I'd probably just leave the free as it was before and call it good so we don't worry about introducing any new bugs, but it's up to you. If you do make any changes, I'd probably lean towards the more granular locking approach.

ianhattendorf (approved):

PR #1998, opened 2024-05-30
Fix build failure on electron 29+

zawata:
bump nan to include this PR:
https://github.com/nodejs/nan/pull/966


gpetrov:
still fails as it needs a bit more bumping to 2.20

needs also the PR https://github.com/nodejs/nan/pull/970

@ianhattendorf

zawata:
fantastic...

I can get a PR up in a few days.

PR #2014, opened 2025-02-03
Fix Github Action workflow

serggl:
### Why

While working on a test patch to `nodegit` I noticed that Github Actions stopped working  due to https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

### What

This PR updates build steps required for Linux platform, and also adds Node 22 to the build matrix. 

Worth mentioning that this also partially rollback changes made with #1860. Not sure if it's still worth testing against Ubuntu 16.04 which has been EOL since 2021 (like 4 years ago), so I switched it to a more recent (but not edge version) 20.04


serggl:
@zawata @ianhattendorf Please review and approve the test run

serggl:
Guess that this error must be fixed (approved) by the maintainers?
```
Error: ERROR: We're doing an SSH key audit.
Reason: unverified automatically (private key found in a public repository)
Please visit https://github.com/settings/keys/29706139 to approve this key so we know it's safe.
```

serggl:
@zawata the updated key will work now, please approve the test run.

Both Linux and Windows tests should pass, as for the MacOS - there's a flaky repository statistics test, which I think isn't related to the test setup itself. On CI it fails randomly with any of Node 18-22. On my local it passes just fine with Node 20.18.0, but not with 20.18.1

zawata:
I actually had no idea the Test workflow was decoding the key (because test/index.js is also decoding the key?). I've never worked with this workflow before and didn't have the time to debug it yesterday. thanks for the help in rewriting these workflows and fixing various issues(like the coverage parsing error)

